PR #11, opened 2016-11-26
General Cleanup

alanmeadows:
* Made values.yaml consistent throughout charts.  Removed any globals
references in subcharts as these are difficult to override.  Only
ports should be in globals to build URLs which can come as part
of a future commit. The hostname endpoint aspect of a service
will come from openstack-base/_hosts.tpl and the port
would come from the chart itself as a global so other charts
can reference the port to build a complete URL.  Putting the
hostnames themselves as globals in individual charts makes it
difficult to make a sweeping top level FQDN change.

* Cleaned up yaml requirements and incorporated a new _common.tpl
that is distributed to all charts to allow common endpoint naming
while still retaining the ability to install individual charts.

* Fixed keystone URL generation during bootstrap as a correct
URL is critical given keystone uses this to construct all
subsequent URLs in the request. Also allow controlling the
default endpoint version and scheme.

* Added missing NAMESPACE declaration to keystone deployment
as this is required for entrypoint to discover resources
not in the 'default' namespace.

* Refactored all nodeSelector values to be consistent throughout
all charts


alanmeadows:
@PiotrProkop may want to take a look at this as it is a pretty big change.  Goal is to keep charts independently installable (with any real dependancies, of course).  I want the uber openstack chart to be a convenience, not a requirement as it is in openstack-helm.  I sense major issues with upgrades via helm with an uber chart.

keystone/templates/db-sync.yaml
```diff
@@ -31,7 +31,7 @@ spec:
       restartPolicy: OnFailure
       containers:
         - name: keystone-init
```
DTadrzak: s/keystone-init/keystone-db-sync/

keystone/templates/db-sync.yaml
```diff
@@ -31,7 +31,7 @@ spec:
...
-          image: {{ .Values.global.deployment.image.db_sync }}
+          image: {{ .Values.images.db_sync }}
           imagePullPolicy: Always
           command:
           - bash
```
DTadrzak: Remove unnecessary environment variables like COMMAND, DEPENDENCY_JOBS, DEPENDENCY_SERVICE from container. We don't need them anymore.

keystone/templates/deployment.yaml
```diff
@@ -15,12 +15,16 @@ spec:
...
+                "name": "NAMESPACE",
+                "value": "{{ .Release.Namespace }}"
+              },
+              {
                 "name": "DEPENDENCY_SERVICE",
                 "value": "mariadb"
               },
               {
                 "name": "DEPENDENCY_JOBS",
-                "value": "mariadb-seed,nova-db-sync"
```
DTadrzak: We need to add keystone-init and keystone-db-sync instead of nova-db-sync.

keystone/templates/deployment.yaml
```diff
@@ -31,10 +35,10 @@ spec:
...
       nodeSelector:
-        {{ .Values.global.deployment.control_node_label }}: enabled
+        {{ .Values.labels.node_selector_key }}: {{ .Values.labels.node_selector_value }}
       containers:
         - name: keystone-api
-          image: {{ .Values.global.deployment.image.api }}
+          image: {{ .Values.images.api }}
           command:
           - bash 
           - /tmp/start.sh
```
DTadrzak: Remove unnecessary environment variables from container section.

keystone/templates/init.yaml
```diff
@@ -30,7 +30,7 @@ spec:
       restartPolicy: OnFailure
       containers:
         - name: keystone-init
-          image: {{ .Values.global.deployment.image.init }}
+          image: {{ .Values.images.init }}
           imagePullPolicy: Always
           env:
             - name: INTERFACE_NAME
```
DTadrzak: Remove unnecessary environment variables from container section.

keystone/templates/init.yaml
```diff
@@ -30,7 +30,7 @@ spec:
       restartPolicy: OnFailure
```
DTadrzak: Remove unnecessary dependency from annotation section. Keystone-init has dependency on keystone-init job :)

mariadb/requirements.yaml
```diff
@@ -0,0 +1,4 @@
+dependencies:
+  - name: openstack-base
+    repository: http://localhost:8879/charts
+    version: 0.1.0
```
DTadrzak: Add missing new line.

memcached/requirements.yaml
```diff
@@ -0,0 +1,4 @@
+dependencies:
+  - name: openstack-base
+    repository: http://localhost:8879/charts
+    version: 0.1.0
```
DTadrzak: Same here

memcached/templates/deployment.yaml
```diff
@@ -10,10 +10,10 @@ spec:
...
       nodeSelector:
-        {{ .Values.global.deployment.control_node_label }}: enabled
+        {{ .Values.labels.node_selector_key }}: {{ .Values.labels.node_selector_value }}
       containers:
         - name: memcached
-          image: {{ .Values.global.deployment.image.memcached }}
+          image: {{ .Values.images.memcached }}
           imagePullPolicy: Always
           env:
             - name: INTERFACE_NAME
```
DTadrzak: I'm not sure whether we need those envs for memcached.

rabbitmq/templates/service.yaml
```diff
@@ -4,6 +4,6 @@ metadata:
   name: rabbitmq
 spec:
   ports:
-  - port: {{.Values.global.rabbitmq_port_public}}
+  - port: {{.Values.network.port.public}}
   selector:
     app: rabbitmq
```
DTadrzak: New line

rabbitmq/requirements.yaml
```diff
@@ -0,0 +1,4 @@
+dependencies:
+  - name: openstack-base
+    repository: http://localhost:8879/charts
+    version: 0.1.0
```
DTadrzak: missing new line

rabbitmq/values.yaml
```diff
@@ -2,16 +2,25 @@
...
+  default_pass: password
+  admin_user: rabbitmq
+  admin_pass: password
+
+network:
+  port:
+    public: '5672'
+    management: '15672'
+
+
```
DTadrzak: Remove a blank line.

alanmeadows:
@DTadrzak issues should be addressed now

keystone/templates/db-sync.yaml
DTadrzak: you change wrong keystone-init string :P, dependency job was corrected, I mean name: keystone-init to name: keystone-db-sync.

rabbitmq/values.yaml
DTadrzak: You forgot about that. Let's stay with 1 blank line between sections.

alanmeadows:
@DTadrzak I think we've finally got all items incorporated.

PR #13, opened 2016-11-28
Add MaaS helm chart

larryrensing:


maas/templates/deployment.yaml
```diff
@@ -0,0 +1,38 @@
...
+  name: maas-region
+spec:
+  template:
+    metadata:
+      labels:
+        app: maas-region
+    spec:
+      containers:
+      - name: maas-region
+        image: {{ .Values.image_maas_region }}
```
DTadrzak: Let's make it consistent to other charts in this repo, so that it would be more readable.
Can you change it to .Values.images.mass_region?

maas/templates/deployment.yaml
```diff
@@ -0,0 +1,38 @@
...
+        imagePullPolicy: Always
+        ports:
+        - containerPort: {{ .Values.maas_region_container_port }}
```
DTadrzak: .Values.network.port.region_container

maas/templates/deployment.yaml
```diff
@@ -0,0 +1,38 @@
...
+        securityContext:
+          privileged: true
+---
```
DTadrzak: Can we split this file into two independent files?

maas/templates/deployment.yaml
```diff
@@ -0,0 +1,38 @@
...
+spec:
+  template:
+    metadata:
+      labels:
+        app: maas-rack-controller
+    spec:
+      hostNetwork: true
+      containers:
+      - name: maas-rack
+        image: {{ .Values.image_maas_rack }}
```
DTadrzak: .Values.images.mass_rack

maas/templates/deployment.yaml
```diff
@@ -0,0 +1,38 @@
...
+        imagePullPolicy: Always
+        ports:
+        - containerPort: {{ .Values.maas_rack_container_port }}
```
DTadrzak: .Values.network.port.rack_container

maas/templates/service.yaml
```diff
@@ -0,0 +1,18 @@
+apiVersion: v1
+kind: Service
+metadata:
+  name: maas-region-ui
+  labels:
+    app: maas-region-ui
+spec:
+  ports:
+  - port: {{ .Values.service_gui_port }}
+    targetPort: {{ .Values.service_gui_target_port }}
```
DTadrzak: .Values.network.port.service_gui_target

maas/templates/service.yaml
```diff
@@ -0,0 +1,18 @@
+apiVersion: v1
+kind: Service
+metadata:
+  name: maas-region-ui
+  labels:
+    app: maas-region-ui
+spec:
+  ports:
+  - port: {{ .Values.service_gui_port }}
```
DTadrzak: .Values.network.port.service_gui

maas/templates/service.yaml
```diff
@@ -0,0 +1,18 @@
...
+    targetPort: {{ .Values.service_gui_target_port }}
+    protocol: TCP
+    name: gui
+  - port: {{ .Values.service_proxy_port }}
+    targetPort: {{ .Values.service_proxy_target_port }}
```
DTadrzak: .Values.network.port.service_proxy_target

maas/templates/service.yaml
```diff
@@ -0,0 +1,18 @@
...
+  name: maas-region-ui
+  labels:
+    app: maas-region-ui
+spec:
+  ports:
+  - port: {{ .Values.service_gui_port }}
+    targetPort: {{ .Values.service_gui_target_port }}
+    protocol: TCP
+    name: gui
+  - port: {{ .Values.service_proxy_port }}
```
DTadrzak: .Values.network.port.service_proxy

maas/templates/deployment.yaml
```diff
@@ -0,0 +1,38 @@
+apiVersion: extensions/v1beta1
```
DTadrzak: You don't need to add node selector for those deployments?

maas/values.yaml
```diff
@@ -0,0 +1,14 @@
+# Default values for maas.
+# This is a YAML-formatted file.
+# Declare variables to be passed into your templates.
+
+image_maas_region: quay.io/attcomdev/maas-region:1.0.1
+image_maas_rack: quay.io/attcomdev/maas-rack:1.0.1
+
+maas_region_container_port: 80
+maas_rack_container_port: 80
```
DTadrzak: you add only service for MaaS_region, is it correct?

maas/values.yaml
```diff
@@ -0,0 +1,14 @@
...
+
+service_gui_port: 80
+service_gui_target_port: 80
+service_proxy_port: 8000
```
DTadrzak: You don't use port 8000 inside container.

alanmeadows:
@DTadrzak Great feedback.  This is why I wanted to start the consistency approach early.  Agreed on all points above.

As a side note, Larry is also working on breaking this up to remove the apt nightmare that is maas and add PVC storage.  We'll also try to incorporate some maas jobs to create the provisioning network.  Right now it is only a handful of post-install steps to have it provision hosts but that is a handful of steps too many.

v1k0d3n:
@DTadrzak makes some really great points. @larryrensing sync up with @alanmeadows (I'm sure you already have) for some clarification on what we're looking for, and then I'm going to let Alan merge these changes/reviews.

maas/values.yaml
alanmeadows: @DTadrzak Just to clarify, assume you mean "You don't use port 80 inside container" - looks like @larryrensing has port 80 exposed on the rack container which needs no port exposure with hostNetworking (and plus, doesn't have a service on port 80 anyways, only DHCP/TFTP and outbound RPC to region controller).

PR #15, opened 2016-11-30
ensure ceph storage is persistent

alanmeadows:


v1k0d3n:
LGMT; good changes. @alanmeadows should we provide a chart for some way to stash/manage secrets? etcd could be leveraged fairly easily/quickly for this task, using k8s third party resource.

PR #16, opened 2016-12-01
Various fixes validating refactored charts

alanmeadows:


alanmeadows:
Note that this pull request adds some environment variables (e.g. to memcache) back in that we removed as part of a cleanup.  These environment vars turn out to still be necessary until we switch to init-containers properly/consistently across all charts.  Will send that separately.

PR #33, opened 2016-12-12
Update readme to fix secret generation ordering

intlabs:
The quickstart guide was attempting to build the helm charts for Ceph before the secrets currently embedded within it were prepared. This commit fixes that and also provides instructions for installing Sigil.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,7 +30,7 @@ Note that in addition to Kubernetes 1.5.0-beta.1, you will need to replace the k
 
 The kubelet should pick up the change and restart the container.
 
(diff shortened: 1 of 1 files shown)


alanmeadows:
LGTM - the Sigil call out was on my list, glad its done.

intlabs:
@alanmeadows would be great to remove the requirement for Sigil at all. I made a quick prototype of a k8s job that performed secret generation for your implementation -  I'd be happy to clean it up and send it over to you if you're interested? I'm also trying to find out if its possible to find a universal way of determining the CIDR for Pods regardless of the CNI plugin used, which would also remove the requirement for manually feeding this in.

alanmeadows:
@intlabs I think that would be great Pete.  The issue with the CIDR requirement is that if its not done correctly, it basically silently implodes and if you don't know what you're looking for it it can be a huge pain to debug so if we can eliminate that I think it would be a great win.

That said, take a look at the latest ceph pull request.  Our other developers are asking to ensure that ceph, and also all the things that depend on ceph for PVCs support bring-your-own-ceph.  We want to make sure whatever we do for convenience now for ceph we can "turn off."

PR #39, opened 2016-12-13
Allow more control over ceph chart

alanmeadows:
* A new boolean value controls whether secrets and
the ceph.conf secret ConfigMaps are installed allowing
you to bring-your-own ceph configs. This should
remove the dependency on having to generate secrets
using the common chart if you have installed your own.

* A new boolean value controls whether we install
a storage class as part of the ceph chart install allowing
you to externally manage this

* More of the ceph values have been converted to values
variables, paving the way for a master yaml to provide input
and overrides to many independent "sub" charts without a parent chart

### DIFF
--- a/ceph/templates/secrets.yaml
+++ b/ceph/templates/secrets.yaml
@@ -1,3 +1,4 @@
+{{- if .Values.secrets.use_common_secrets -}}
 ---
 apiVersion: v1
 kind: Secret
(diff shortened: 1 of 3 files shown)


v1k0d3n:
This looks awesome @alanmeadows; much cleaner. I think @portdirect will be interested in this also.

PR #50, opened 2016-12-15
Fix mariadb-deployment.yaml file

DTadrzak:
According to
* http://kubernetes.io/docs/api-reference/extensions/v1beta1/definitions/#_v1_podtemplatespec
* http://kubernetes.io/docs/api-reference/extensions/v1beta1/definitions/#_v1_persistentvolumeclaimvolumesource

SecurityContext is placed in wrong path. What's more persistentVolumeClaim
doesn't support matchLabels.

---
This change is [Reviewable](https://reviewable.io/reviews/att-comdev/aic-helm/50)

### DIFF
--- a/mariadb/templates/mariadb-deployment.yaml
+++ b/mariadb/templates/mariadb-deployment.yaml
@@ -41,8 +41,6 @@ metadata:
 spec:
   replicas: 1
   template:
-    securityContext:
-      runAsUser: 0      
     metadata:
       name: mariadb-{{$v}}
       labels:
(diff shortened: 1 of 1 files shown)


alanmeadows:
@DTadrzak with whats in PR#49, is this necessary? The StatefulSet version doesn't leverage the security context functionality and I have not seen any resulting issues from that removal during testing.  Also, the volume label match syntax is similarly removed as it is replaced with a volumeClaimTemplate which is a lot cleaner.

alanmeadows:
@DTadrzak I think we can close this since we are no longer escalating the mariadb deployment:

https://github.com/att-comdev/aic-helm/blob/master/mariadb/templates/deployment.yaml

PR #52, opened 2016-12-16
Glance chart bugfixes

alanmeadows:
* imagePullPolicy requirements for init-containers as required by helm 2.1.0 which does not supply this by default

* dependency tree specified mariadb-init, when there is only a
mariadb-seed job to depend on

* the requirements.yaml should not include any chart in this
repository other then common as that severely complicates
removing charts as all dependent elements are removed with it

* the post.sh.yaml has HOME set to /tmp which will not read
/home/ansible configuration.  It was unclear if this was by
design, but /home/ansible seems like an important part of the
kolla toolbox

* the post.sh.yaml file had quoting typos, but even when they are
fixed the job/glance-post will not run to completion, complaining
of a missing kolla_keystone_service module

---
This change is [Reviewable](https://reviewable.io/reviews/att-comdev/aic-helm/52)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,10 +3,10 @@
 B64_DIRS := common/secrets
 B64_EXCLUDE := $(wildcard common/secrets/*.b64)
 
-CHARTS := ceph mariadb rabbitmq memcached keystone openstack
+CHARTS := ceph mariadb rabbitmq GLANCE memcached keystone glance openstack
 COMMON_TPL := common/templates/_globals.tpl
 
-all: common bootstrap mariadb rabbitmq memcached keystone openstack
+all: common bootstrap mariadb rabbitmq memcached keystone glance openstack
 
 common: build-common
 
@@ -19,6 +19,8 @@ mariadb: build-mariadb
 keystone: build-keystone
 
 rabbitmq: build-rabbitmq
+
+glance: build-glance
 
 memcached: build-memcached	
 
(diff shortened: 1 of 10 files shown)


alanmeadows:
@PiotrProkop may want to see these changes.  Please note that jobs/glance-post still does not run to completion due to the issue mentioned, namely:

```
$ ok logs po/glance-post-8zs33 -f
+ ansible localhost -vvv -m kolla_keystone_service -a 'service_name=glance \
service_type=image \
description="Openstack Image" \
endpoint_region="RegionOne" \
url="http://glance-api:9292" \
interface=admin \
region_name="RegionOne" \
auth="{# openstack_glance_auth #}"' -e '{ '\''openstack_glance_auth'\'': {'\''auth_url'\'':'\''http://keystone-api:35357'\'', '\''username'\'':'\''admin'\'','\''password'\'':'\''password'\'','\''project_name'\'':'\''admin'\'','\''domain_name'\'':'\''default'\''} }'
[WARNING]: log file at /var/log/kolla/ansible.log is not writeable and we cannot create it, aborting

Using /home/ansible/.ansible.cfg as config file
localhost | FAILED! => {
    "failed": true, 
    "msg": "The module kolla_keystone_service was not found in configured module paths. Additionally, core modules are missing. If this is a checkout, run 'git submodule update --init --recursive' to correct this problem."
}
```

alanmeadows:
This chart still requires a refactor to bring it inline with keystone and now mariadb.

PR #53, opened 2016-12-16
Keystone chart bugfixes

alanmeadows:
* start.sh was added back, which had required env sourcing for successful apache startup.

* the naming convention for charts is finalized with this example
landing on configmap-*.yaml to satisfy those of us with OCD

* imagePullPolicies added for init-containers, required by
helm 2.1.0 which does not supply them by default

---
This change is [Reviewable](https://reviewable.io/reviews/att-comdev/aic-helm/53)

### DIFF
--- a/keystone/values.yaml
+++ b/keystone/values.yaml
@@ -14,6 +14,7 @@ images:
   api: quay.io/stackanetes/stackanetes-keystone-api:newton
   init: quay.io/stackanetes/stackanetes-kolla-toolbox:newton
   entrypoint: quay.io/stackanetes/kubernetes-entrypoint:v0.1.0
+  pull_policy: "IfNotPresent"
 
 keystone:
   version: v2.0
--- a/common/templates/_hosts.tpl
+++ b/common/templates/_hosts.tpl
@@ -8,7 +8,7 @@
 {{- define "mariadb_host"}}mariadb.{{.Release.Namespace}}.svc.{{ include "region" . }}.{{ include "tld" . }}{{- end}}
 
 # keystone
-{{- define "keystone_db_host"}}{{ include "mariadb_host" . }}{{end}}
+{{- define "keystone_db_host"}}{{ include "mariadb_host" . }}{{- end}}
 {{- define "keystone_api_endpoint_host_admin"}}keystone-api.{{.Release.Namespace}}.svc.{{ include "region" . }}.{{ include "tld" . }}{{- end}}
 {{- define "keystone_api_endpoint_host_internal"}}keystone-api.{{.Release.Namespace}}.svc.{{ include "region" . }}.{{ include "tld" . }}{{- end}}
 {{- define "keystone_api_endpoint_host_public"}}keystone-api.{{ include "region" . }}.{{ include "tld" . }}{{- end}}
--- a/keystone/templates/bin/_start.sh.tpl
+++ b/keystone/templates/bin/_start.sh.tpl
@@ -1,0 +1,8 @@
+#!/bin/bash		
+set -ex		
+		
+# Loading Apache2 ENV variables		
+source /etc/apache2/envvars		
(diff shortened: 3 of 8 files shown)


alanmeadows:
@DTadrzak please see -- this addresses issue raised in commit https://github.com/att-comdev/aic-helm/commit/3581d443a42fb716a34a5bac270d546e51cb8959

v1k0d3n:
Review status: 0 of 8 files reviewed at latest revision, 2 unresolved discussions.

---

*[keystone/templates/deployment.yaml, line 47 at r2](https://reviewable.io:443/reviews/att-comdev/aic-helm/53#-KZD4UfkkhJCv0zt3wpl:-KZD4Ufl1LHJ5oWCX50O:bs26gt9) ([raw file](https://github.com/att-comdev/aic-helm/blob/ff61c50dcc15e855c73ae6be0d711d0ccc7c5ee4/keystone/templates/deployment.yaml#L47)):*
> ```YAML
>         - name: keystone-api
>           image: {{ .Values.images.api }}
>           imagePullPolicy: {{ .Values.images.pull_policy }}
> ```

definitely needed for any of our charts/manifests for k8s1.5+/helm2.1.0+. this change has been disruptive for charts in k8s, but this is obviously k8s path going forward.

---

*[keystone/templates/deployment.yaml, line 48 at r2](https://reviewable.io:443/reviews/att-comdev/aic-helm/53#-KZD4b3OoJSYkQTeO0tJ:-KZD4b3P9xV_2ans-zKu:b-73w2bp) ([raw file](https://github.com/att-comdev/aic-helm/blob/ff61c50dcc15e855c73ae6be0d711d0ccc7c5ee4/keystone/templates/deployment.yaml#L48)):*
> ```YAML
>           image: {{ .Values.images.api }}
>           imagePullPolicy: {{ .Values.images.pull_policy }}
>           command:
> ```

really like the removal of the forced dependency. i see the need to be ultra-flexible.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/att-comdev/aic-helm/53)*

keystone/values.yaml
```diff
@@ -14,6 +14,7 @@ images:
   api: quay.io/stackanetes/stackanetes-keystone-api:newton
   init: quay.io/stackanetes/stackanetes-kolla-toolbox:newton
   entrypoint: quay.io/stackanetes/kubernetes-entrypoint:v0.1.0
+  pull_policy: "IfNotPresent"
```
DTadrzak: Let's move it as a common variable.

v1k0d3n: good idea. as long as users can continue to override at an atomic level...that should be fine.

alanmeadows:
If we establish and simply reference this as a global value, that will make
overriding it on a per chart difficult at least with the values.py plan we
have today.

It's already a little aggressive in scope as it applies to all images in
the keystone chart so I don't want to expand that even further.

What we may do instead to account for your goal is set this to nil in the
chart and use | default .Values.global.images.pull_policy.  That way in an
environmental config churned out by values.py you could set something more
granular for just keystone and glance, for instance but let's do that in a
separate commit.

Alan Meadows

I suggest we let this in as is for now and come up

v1k0d3n:
Reviewed 9 of 10 files at r1, 1 of 1 files at r2.
Review status: all files reviewed at latest revision, 3 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/att-comdev/aic-helm/53)*

v1k0d3n:
:lgtm:

---

Review status: all files reviewed at latest revision, 3 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/att-comdev/aic-helm/53#-:-KZRjqzDTEAwT3R0jwN6:bnfp4nl)*

PR #62, opened 2016-12-20
Initial commit of horizon chart

alanmeadows:
This change is [Reviewable](https://reviewable.io/reviews/att-comdev/aic-helm/62)


horizon/templates/deployment.yaml
```diff
@@ -0,0 +1,69 @@
...
+            - name: startsh
+              mountPath: /tmp/start.sh
+              subPath: start.sh
+            - name: horizonconf
+              mountPath: /etc/apache2/sites-enabled/000-default.conf
+              subPath: horizon.conf
+            - name: localsettings
+              mountPath: /etc/openstack-dashboard/local_settings
+              subPath: local_settings
+          securityContext:
```
DTadrzak: I think that we shouldn't run container as root if we don't need it.

alanmeadows: I agree _completely_ on principal. 

In fact, I think we need to make a complete second pass on all charts for security enhancements.  Secondly, I think we need to make a revisit across all charts for shell script cleanup as many of them are rather curt or in some places a little hackish.  Our timeline constraints mean we need to achieve something working by early January, but we need to quickly change gears and start operationalizing this after that and making sure every shell script is consistent and clean.  I think some of what I see in fuel-ccp in terms of shell code is ripe for plucking, which means we now have three sources to pull from: stackanetes, sap openstack-helm, and fuel-ccp.

To be sure, anytime there is a call to run a container as root at least from me it is because it was required without refactoring the approach. Since this chartis based on stackanetes, through and through, I was unable to get the horizon service to launch without that as the CWD for apache was the python environment directory, which needed non-horizon user privileges to generate temporary files.

We could either refactor the launch here so the privilege can be removed or fix this as part of a security sweep post January 10 but I am not sure how the original stackanetes launch of this in its non-chart was working without either escalated privileges, or a change to how horizon is launched to fix the need for extra privileges.

horizon/templates/deployment.yaml
```diff
@@ -0,0 +1,69 @@
...
+            "image": "{{ .Values.images.entrypoint }}",
+            "imagePullPolicy": "{{ .Values.images.pull_policy }}",
+            "env": [
+              {
+                "name": "NAMESPACE",
+                "value": "{{ .Release.Namespace }}"
+              },
+              {
+                "name": "DEPENDENCY_SERVICE",
+                "value": "memcached,keystone-api"
```
alanmeadows: This likely needs to follow the new approach of a valued based dependency breakout.

horizon/values.yaml
```diff
@@ -0,0 +1,36 @@
...
+  horizon_secret_key: 9aee62c0-5253-4a86-b189-e0fb71fa503c
+  debug: "True"
+
+# typically overriden by environmental 
+# values, but should include all endpoints
+# required by this chart
+endpoints:
+  keystone:
+    hosts:
+      default: keystone-api.openstack.cluster.local
```
alanmeadows: I think this should simply be keystone-api to allow the current approach but remain agnostic of the namespace used

alanmeadows:
@v1k0d3n @PiotrProkop @DTadrzak this configmap annotation approach in https://github.com/att-comdev/aic-helm/pull/62/commits/ca2dd29bbd88f00e03b4ee49d713aa28bb17426e included in this PR is what I would also like to apply to other charts, to support rolling updates on configmap changes (conf file tweaks) without inhibiting helm rollback functionality - any feedback welcome.

Long term, it might be nice to exclude certain files in the configmap as not requiring restarts, but I do not want to go back to one-file-one-configmap.  What would fit into this is a configmap-norestart.yaml to house those which is simply not annotated.

v1k0d3n:
Reviewed 11 of 15 files at r1, 3 of 3 files at r2, 1 of 1 files at r3.
Review status: 15 of 16 files reviewed at latest revision, 4 unresolved discussions.

---

*[horizon/requirements.lock, line 6 at r3](https://reviewable.io:443/reviews/att-comdev/openstack-helm/62#-KZihYgdeq8b11jfoR8z:-KZihYgdeq8b11jfoR9-:b-cvn961) ([raw file](https://github.com/att-comdev/openstack-helm/blob/1c6fc76842b277d8398ece26d344afcbd5f6caad/horizon/requirements.lock#L6)):*
> ```
>   version: 0.1.0
> digest: sha256:c6a7e430c900036912fe3fdc14213e9280c5da0b6607ce4dcf6dc95535d114fc
> generated: 2016-11-30T17:10:50.917838584-08:00
> ```

the update in .gitignore should prevent these `*.lock` files, but you may want to remove it from the commit.

---

*[horizon/templates/deployment.yaml, line 58 at r1](https://reviewable.io:443/reviews/att-comdev/openstack-helm/62#-KZRszBUa8sXYk03K3Xs-r1-58:-KZin1ZwUcNAd-8ItdSq:b-p1lesv) ([raw file](https://github.com/att-comdev/openstack-helm/blob/6ec92626e854fabdf234ddcc57fb9a534e217e57/horizon/templates/deployment.yaml#L58)):*
<details><summary><i>Previously, alanmeadows (Alan Meadows) wrote…</i></summary><blockquote>

I agree _completely_ on principal. 

In fact, I think we need to make a complete second pass on all charts for security enhancements.  Secondly, I think we need to make a revisit across all charts for shell script cleanup as many of them are rather curt or in some places a little hackish.  Our timeline constraints mean we need to achieve something working by early January, but we need to quickly change gears and start operationalizing this after that and making sure every shell script is consistent and clean.  I think some of what I see in fuel-ccp in terms of shell code is ripe for plucking, which means we now have three sources to pull from: stackanetes, sap openstack-helm, and fuel-ccp.

To be sure, anytime there is a call to run a container as root at least from me it is because it was required without refactoring the approach. Since this chartis based on stackanetes, through and through, I was unable to get the horizon service to launch without that as the CWD for apache was the python environment directory, which needed non-horizon user privileges to generate temporary files.

We could either refactor the launch here so the privilege can be removed or fix this as part of a security sweep post January 10 but I am not sure how the original stackanetes launch of this in its non-chart form was working without either escalated privileges, or a change to how horizon is launched to fix the need for extra privileges.
</blockquote></details>

What are your thoughts on tracking this via an issue? I think there are a couple of approaches here, but if we move it to an issue we can track/link the discussion for more folks to participate in. Let's track this as a second release improvement.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/att-comdev/openstack-helm/62)*

alanmeadows:
Review status: 15 of 16 files reviewed at latest revision, 4 unresolved discussions.

---

*[horizon/requirements.lock, line 6 at r3](https://reviewable.io:443/reviews/att-comdev/openstack-helm/62#-KZihYgdeq8b11jfoR8z:-KZitp5JSkUGF_N5cwvw:b9ppnsd) ([raw file](https://github.com/att-comdev/openstack-helm/blob/1c6fc76842b277d8398ece26d344afcbd5f6caad/horizon/requirements.lock#L6)):*
<details><summary><i>Previously, v1k0d3n (Brandon B. Jozsa) wrote…</i></summary><blockquote>

the update in .gitignore should prevent these `*.lock` files, but you may want to remove it from the commit.
</blockquote></details>

Nice catch, will remove.

---

*[horizon/templates/deployment.yaml, line 58 at r1](https://reviewable.io:443/reviews/att-comdev/openstack-helm/62#-KZRszBUa8sXYk03K3Xs-r1-58:-KZithfXyEOtgVO-WxOl:b-pbdqz3) ([raw file](https://github.com/att-comdev/openstack-helm/blob/6ec92626e854fabdf234ddcc57fb9a534e217e57/horizon/templates/deployment.yaml#L58)):*
<details><summary><i>Previously, v1k0d3n (Brandon B. Jozsa) wrote…</i></summary><blockquote>

What are your thoughts on tracking this via an issue? I think there are a couple of approaches here, but if we move it to an issue we can track/link the discussion for more folks to participate in. Let's track this as a second release improvement.
</blockquote></details>

Agreed.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/att-comdev/openstack-helm/62)*

PR #63, opened 2016-12-20
Another Glance-fix.

DTadrzak:
Without those changes i wasn't able to run glance-post.
Necessary changes in order to make glance-post work.

---
This change is [Reviewable](https://reviewable.io/reviews/att-comdev/aic-helm/63)

### DIFF
--- a/glance/templates/post.yaml
+++ b/glance/templates/post.yaml
@@ -44,6 +44,9 @@ spec:
             - name: postsh
               mountPath: /tmp/post.sh
               subPath: post.sh
+          env:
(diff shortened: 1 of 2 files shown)


glance/templates/post.yaml
```diff
@@ -44,6 +44,9 @@ spec:
             - name: postsh
               mountPath: /tmp/post.sh
               subPath: post.sh
+          env:
```
alanmeadows: I think this is the core thing that was originally missing, in fact I noticed it in your WIP nova commit.  Good to know this is a requirement for toolbox work.

PR #65, opened 2016-12-23
general bugfixes and doc updates

v1k0d3n:
This patch fixes several bugs that were uncovered while redrafting our docs. It provides the following:
* new docs folder along with getting-started guides (wip)
* bugfix: resolve crashloop for rgw deployment. it is not required for this type of deployment, although we've allowed the flexibility to override this in the `./ceph/values.yaml` declaration.
* bugfix: cleaned up "nil" statement in `./ceph/values.yaml`, however this causes issues in the linting process for helm during upgrades. so as a temporary workaround...
* workaround: we created a static/hardcoded namespace declaration in `./ceph/templates/storage.yaml`.

Notes: for our workaround, Helm core team should be engaged to determine if we are using incorrectly or if this is a bug that we've uncovered. Either way, and issue should be created for this, and eventually the hardcoded declaration should be removed.

---
This change is [Reviewable](https://reviewable.io/reviews/att-comdev/openstack-helm/65)


docs/installation/getting-started.md
```diff
@@ -0,0 +1,249 @@
...
+If any helm install step fails, you can back it out with ```helm delete --purge <releaseName>```
+
+Make sure sigil is installed to perform the ceph secret generation, as noted in the [Requirements](Requirements).
+
+```
+# label all known nodes as candidates for pods
+kubectl label nodes openstack-control-plane=enabled --all
+kubectl label nodes ceph-storage=enabled --all
+
+# move into the aic-helm directory
```
alanmeadows: Don't think you meant aic-helm here.

docs/installation/getting-started.md
```diff
@@ -0,0 +1,249 @@
...
+cd aic-helm
```
alanmeadows: Again here.

docs/installation/getting-started.md
```diff
@@ -0,0 +1,249 @@
...
+# create some local directories used as nodeDirs
+# for persistent storage
+mkdir -p /var/lib/openstack-helm/ceph
+
+# generate secrets (ceph, etc.)
+cd common/utils/secret-generator
+./generate_secrets.sh all `./generate_secrets.sh fsid`
+cd ../../..
+
+# now you are ready to build aic-helm
```
alanmeadows: And here.

PR #71, opened 2016-12-29
Add development mode for mariadb chart

alanmeadows:
This can be enabled with:

helm install --name mariadb --set development.enabled=true local/mariadb

This will ensure only a single replica is used and that PVCs (which
require a persistent volume solution) like ceph are not used.  Instead
this flag enables a single replica with hostDir storage using the
host path defined in values.yaml which defaults to the following:

/var/lib/openstack-helm/mariadb

---
This change is [Reviewable](https://reviewable.io/reviews/att-comdev/openstack-helm/71)

### DIFF
--- a/mariadb/templates/bin/_seed.sh.tpl
+++ b/mariadb/templates/bin/_seed.sh.tpl
@@ -57,7 +57,11 @@ function wait_for_cluster {
 # the implementation will be switched to Deployment
 # (using anti-affinity feature).
 
+{{- if .Values.development.enabled }}
+REPLICAS=1
+{{- else }}
 REPLICAS={{ .Values.replicas }}
+{{- end }}
 
 if [ "$REPLICAS" -eq 1 ] ; then
     echo "Requested to build one-instance MariaDB cluster. There is no need to run seed. Exiting."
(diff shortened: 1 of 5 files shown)


v1k0d3n:
I will test, and when completed (with docs) will merge in these changes.

v1k0d3n:
Review status: 0 of 5 files reviewed at latest revision, 1 unresolved discussion.

---

*[mariadb/values.yaml, line 22 at r2](https://reviewable.io:443/reviews/att-comdev/openstack-helm/71#-K_CCjEFcWK-ukl_i7Ia:-K_CCjEG0OcXYyY0JEww:bbhnejq) ([raw file](https://github.com/att-comdev/openstack-helm/blob/fe8ad47f9b851d14ab59fbb58d940d94b1294f06/mariadb/values.yaml#L22)):*
> ```YAML
> development:
>   enabled: false
>   storage_path: /var/lib/openstack-helm/mariadb
> ```

Can you change this to the following?

`/data/openstack-helm/mariadb`

This is per: https://github.com/kubernetes/minikube#persistent-volumes

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/att-comdev/openstack-helm/71)*

v1k0d3n:
going to merge and then change with docs. easy enough to update at the same time.

PR #77, opened 2017-01-02
OpenStack Heat Initial Commit

intlabs:
Initial commit of Heat Chart

Until Neutron & Nova are in place it is not possible to test this chart fully. The service endpoints will require work to communicate with VMs, which depends on the Neutron deployment arch and how we plan to expose services externally. My experience is that external service exposure is probably best handled via ingress controllers at the edge of the control-plane nodes, although there are other methods of achieving a similar effect.

The config file layout/arch in this PR uses a similar methodology to the one I used in Harbor.
User/service and endpoint generation is performed by the OpenStack CLI, which removes the dependency on Kolla's ansible plugins and increases flexibility.

---
This change is [Reviewable](https://reviewable.io/reviews/att-comdev/openstack-helm/77)


heat/templates/jobs/keystone/service/bin/_ks-service.sh.tpl
```diff
@@ -0,0 +1,35 @@
...
+OS_SERVICE_DESC="${OS_REGION_NAME}: ${OS_SERVICE_NAME} (${OS_SERVICE_TYPE}) service"
+
+# Get Service ID if it exists
+unset OS_SERVICE_ID
+OS_SERVICE_ID=$( openstack service list -f csv --quote none | \
+                 grep ",${OS_SERVICE_NAME},${OS_SERVICE_TYPE}$" | \
+                 sed -e "s/,${OS_SERVICE_NAME},${OS_SERVICE_TYPE}//g" )
+
+# If a Service ID was not found, then create the service
+if [[ -z ${OS_SERVICE_ID} ]]; then
```
alanmeadows: One of the things I found lacking with the kolla ansible foo (unless it was user error) that I enjoyed so much with the chef work I used to participate in was the ability to update keystone definitions for services.

I'd like for us to start thinking about having a consistently named job in each chart: register.

The purpose of breaking it out is you can control just the registration aspect, either by deleting the job so its re-run on helm upgrade ensuring you do not run any other debilitating/resetting init functions.

If we for instance wanted to update either how the endpoint was inserted (my fqdn changed, my internal IPs changed, etc.) or the service description formed (the first is a real world example, the second just for illustration), you would imagine we could delete the jobs/heat-register job allowing it to re-run. Perhaps helm could provide a flag for this ensuring its re-run or theres some more graceful way to trigger its re-run then a delete.

In any case, what it depends on though at its core is all of our scripts being able to update keystone definitions for their services.  Today, all they know how to do is insert and if the newly calculated insert is different then the one that exists, too bad.

It may be a step forward toward this if this script supports updating the service and endpoints since it seems to me you are building an alternate registration script set that can be generically used.

And finally, toward that end, do these perhaps belong as common scripts for all scripts to leverage if they want to get off kolla specific registration scripts?

heat/templates/jobs/keystone/endpoints/ks-endpoints.yaml
```diff
@@ -0,0 +1,96 @@
...
+                "value": "echo done"
+              }
+            ]
+          }
+        ]'
+    spec:
+      restartPolicy: OnFailure
+      containers:
+        - name: heat-ks-endpoints-admin
+{{ include "container_ks_endpoint" . | indent 10 }}
```
alanmeadows: It feels like thee is a weird mix of include and static data here forming the whole.

Perhaps this is a case for the define to be like the "template" definition in common, which supports an incoming tuple?

That way there is a single include with input for all the service specific params. 

It would be nice if a there was a better pattern then just an unnamed tuple for input but it is the only one I am familiar with right now.

{{ tuple "admin" "heat" ... . | include "container_ks_endpoint" | indent ... }}

heat/templates/jobs/keystone/endpoints/ks-endpoints.yaml
```diff
@@ -0,0 +1,96 @@
...
+            - name: OS_SVC_ENDPOINT
+              value: admin
+            - name: OS_SERVICE_NAME
+              value: heat
+            - name: OS_SERVICE_TYPE
+              value: orchestration
+            - name: OS_SERVICE_ENDPOINT
+              value: {{ .Values.service.api.proto }}://{{ .Values.service.api.name }}:{{ .Values.service.api.port }}/v1/\$(project_id)s
```
alanmeadows: It is my fault for not putting this up, but I have been doing this:

a nova file requiring an endpoint for nova (on internal endpoint):

```
url='{{ include "endpoint_nova_api_internal" . }}' \
```

the nova values.py:

```
# typically overriden by environmental
# values, but should include all endpoints
# required by this chart
endpoints:
  glance:
    hosts:
      default: glance-api
    type: image
    path: null
    scheme: 'http'
    port:
      api: 9292
      registry: 9191
  nova:
    hosts:
      default: nova-api
    path: "/v2/%(tenant_id)s"
    type: compute
    scheme: 'http'
    port:
      api: 8774
      metadata: 8775
      novncproxy: 6080
  keystone:
    hosts:
      default: keystone-api
    path: /v3
    type: identity
    scheme: 'http'
    port:
        admin: 35357
        public: 5000
```

common/templates/_endpoints.tpl supporting this lookup:

```
{{- define "endpoint_nova_api_internal" -}}
{{- $fqdn := .Release.Namespace -}}
{{- if .Values.endpoints.fqdn -}}
{{- $fqdn := .Values.endpoints.fqdn -}}
{{- end -}}
{{- with .Values.endpoints.nova -}}
        {{.scheme}}://{{.hosts.internal | default .hosts.default}}.{{ $fqdn }}:{{.port.api}}{{.path}}
{{- end -}}
{{- end -}}
```

alanmeadows: Open to comments on this approach.

intlabs:
Review status: 0 of 70 files reviewed at latest revision, 7 unresolved discussions.

---

*[heat/templates/_helpers.tpl, line 1 at r1](https://reviewable.io:443/reviews/att-comdev/openstack-helm/77#-K_Ri1AOBXpfxoNj9iIV-r1-1:-K_Vy5fO7Up226HJlAZa:b-plri5j) ([raw file](https://github.com/att-comdev/openstack-helm/blob/f02e0eb78dcee498c6522dd1e21c2b98d66799dd/heat/templates/_helpers.tpl#L1)):*
<details><summary><i>Previously, alanmeadows (Alan Meadows) wrote…</i></summary><blockquote>

I've moved this to common/templates/_funcs.tpl in my uncommitted nova/neutron work.  I'm not sure why "_helpers.tpl" was spawned with common definitions as defining them in common and using them in the "parent" has been working for me reliably post 2.0.0.  The heat helpers above, specific functions to help with heat makes sense to me to be localized.  

If you want you can move these to common as part of this commit if you could envision them as truly common.  JoinListWithColon (which is humorous because its really joinListWithComma) is definitely used by many charts at this point -- refactoring them to point to common we can do as something entirely separate.
</blockquote></details>

Yeah `JoinListWithColon` made me raise an eyebrow as well, though I thought it would be best to hit this all at once - as I figured that this would be consolidated into a common functions at some point.

---

*[heat/templates/config/heat-api-paste.ini.yaml, line 1 at r1](https://reviewable.io:443/reviews/att-comdev/openstack-helm/77#-K_Ri1AOBXpfxoNj9iIl-r1-1:-K_VyZvKHMTLFdwVf7MZ:bmqqb4d) ([raw file](https://github.com/att-comdev/openstack-helm/blob/f02e0eb78dcee498c6522dd1e21c2b98d66799dd/heat/templates/config/heat-api-paste.ini.yaml#L1)):*
<details><summary><i>Previously, alanmeadows (Alan Meadows) wrote…</i></summary><blockquote>

Pete, notice as in keystone, to avoid the file sprawl, we've been leveraging a "configmap-bin" and "confgimap-etc" .yaml files with definitions for all the relevant script vs config files.  This reduces the number of files where before it was 1 config map per file, its 2 config map files per all files.  As long as subPath is used in the actual mount, it functions virtually the same.
</blockquote></details>

Agree - the approach you're moving toward is nicer - and I'll make the same transition. My only concern is with updates of config maps (though I need to get my head round the hashing annotation stuff I've seen in a few places and notice that you've adopted for Horizon, as I'm struggling to understand how it works beyond a single update of the map - though I must confess I suspect this is because I'm jet to play with it), with a single config map for a service will the update of one component (eg number of API workers) cause all pods/deployments etc that consume that config map be restarted, regardless of whether they consume the subpath in question? If so is this something we should address, by producing a common configmap-bin, and as small a number of component specific ones as possible, or just accept?

---

*[heat/templates/jobs/db/init/bin/_db-init.sh.tpl, line 5 at r1](https://reviewable.io:443/reviews/att-comdev/openstack-helm/77#-K_Ri1APyHbqCBY9ehpF-r1-5:-K_W-RsKlD-vz1OM7p9o:bijtscr) ([raw file](https://github.com/att-comdev/openstack-helm/blob/f02e0eb78dcee498c6522dd1e21c2b98d66799dd/heat/templates/jobs/db/init/bin/_db-init.sh.tpl#L5)):*
<details><summary><i>Previously, alanmeadows (Alan Meadows) wrote…</i></summary><blockquote>

Can we break this up into \ breaks for readability?
</blockquote></details>

Yes - will do!

I'm not the most familiar with the mysql_db module in ansible, and will need to research it more - but I was wondering about changing this to use the MySQL client in another PR? I'm using PKI/TLS for database's currently and would like to be able to support that here - and am not sure that ansible can cope with this.

---

*[heat/templates/jobs/keystone/endpoints/ks-endpoints.yaml, line 34 at r1](https://reviewable.io:443/reviews/att-comdev/openstack-helm/77#-K_Ri1APyHbqCBY9ehpN-r1-34:-K_W-ZjFUugilI7Un574:b-5s8bvd) ([raw file](https://github.com/att-comdev/openstack-helm/blob/f02e0eb78dcee498c6522dd1e21c2b98d66799dd/heat/templates/jobs/keystone/endpoints/ks-endpoints.yaml#L34)):*
<details><summary><i>Previously, alanmeadows (Alan Meadows) wrote…</i></summary><blockquote>

It feels like thee is a weird mix of include and static data here forming the whole.

Perhaps this is a case for the define to be like the "template" definition in common, which supports an incoming tuple?

That way there is a single include with input for all the service specific params. 

It would be nice if a there was a better pattern then just an unnamed tuple for input but it is the only one I am familiar with right now.

{{ tuple "admin" "heat" ... . | include "container_ks_endpoint" | indent ... }}

</blockquote></details>

Yeah - I wasn't sure about this either but thought we should get the conversation started somewhere :)

I'll have a look in-to the options for making this cleaner, and get them in the next PS for review.

---

*[heat/templates/jobs/keystone/endpoints/ks-endpoints.yaml, line 42 at r1](https://reviewable.io:443/reviews/att-comdev/openstack-helm/77#-K_Ri1APyHbqCBY9ehpN-r1-42:-K_W05UgFzAFjQWHpBMM:b4gcfpa) ([raw file](https://github.com/att-comdev/openstack-helm/blob/f02e0eb78dcee498c6522dd1e21c2b98d66799dd/heat/templates/jobs/keystone/endpoints/ks-endpoints.yaml#L42)):*
<details><summary><i>Previously, alanmeadows (Alan Meadows) wrote…</i></summary><blockquote>

Open to comments on this approach.
</blockquote></details>

Looks like a great starting point! I was wondering about this, as it seemed the weakest area in the repo by far. Also looks like it would/will make defining ingress controllers a lot simpler - as we will be able to use the `path` element for forwarding etc.

I'd personally much prefer splitting each (sub?) service for a service (eg `api`, `metadata` and `novncproxy`) into their own service, for two reasons: firstly it makes it easier to control a services specific requirements eg, nova-metadata is a bit special in that its only consumed by VMs and many Neutron SDN backends want to do fun things with it - meaning it's often best running as a headless service, also PKI auth becomes a lot easier to control.

I assume that this example is only for internal? We should provide the same level of control for public endpoints as well, which may be totally different from the ones within the cluster - we may need to get *slightly* opinionated on this front as OpenStack is trying to make the shift toward using `domain.tld/service/service-path` rather than `domain.tld:service-port/service-path` for public endpoints, I'd also like to support `service.domain.tld/service-path` ideally, though it may provide difficult to support all options at first? 

---

*[heat/templates/jobs/keystone/service/bin/_ks-service.sh.tpl, line 29 at r1](https://reviewable.io:443/reviews/att-comdev/openstack-helm/77#-K_Ri1APyHbqCBY9ehpO-r1-29:-K_W4BY0GBrfVKCyGPV0:bf862rm) ([raw file](https://github.com/att-comdev/openstack-helm/blob/f02e0eb78dcee498c6522dd1e21c2b98d66799dd/heat/templates/jobs/keystone/service/bin/_ks-service.sh.tpl#L29)):*
<details><summary><i>Previously, alanmeadows (Alan Meadows) wrote…</i></summary><blockquote>

One of the things I found lacking with the kolla ansible foo (unless it was user error) that I enjoyed so much with the chef work I used to participate in was the ability to update keystone definitions for services.

I'd like for us to start thinking about having a consistently named job in each chart: register.

The purpose of breaking it out is you can control just the registration aspect, either by deleting the job so its re-run on helm upgrade ensuring you do not run any other debilitating/resetting init functions.

If we for instance wanted to update either how the endpoint was inserted (my fqdn changed, my internal IPs changed, etc.) or the service description formed (the first is a real world example, the second just for illustration), you would imagine we could delete the jobs/heat-register job allowing it to re-run. Perhaps helm could provide a flag for this ensuring its re-run or theres some more graceful way to trigger its re-run then a delete.

In any case, what it depends on though at its core is all of our scripts being able to update keystone definitions for their services.  Today, all they know how to do is insert and if the newly calculated insert is different then the one that exists, too bad.

It may be a step forward toward this if this script supports updating the service and endpoints since it seems to me you are building an alternate registration script set that can be generically used.

And finally, toward that end, do these perhaps belong as common scripts for all scripts to leverage if they want to get off kolla specific registration scripts?
</blockquote></details>

Yeah - this, and the endpoints, script are both taken from Harbor which can accommodate the endpoint update example you have above, though could always benefit from more eyes and care :)

I think we need to to find out how to rerun jobs in K8s properly, and see if we can get upstream to help, as far as I know deleting and re-running is the only current option - which is not that nice, I'd kill (slight exaggeration) for a `kubectl job rerun foo`/`kubectl job increment-completions foo` or similar, though perhaps a helm hook may be the answer?

I think having these as common scripts for all services to share makes total sense, and as part of cleaning the yaml up I'll try and work out a way of doing that cleanly, for review in the next commit.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/att-comdev/openstack-helm/77)*

heat/templates/jobs/keystone/endpoints/ks-endpoints.yaml
intlabs: @alanmeadows, sorry I'm not using reviewable.io to update my response, in addition to the comments I made there - I think it is essential that we support (and I hope encourage) people to deploy services to separate namespaces, primarily to improve security but there are also other benefits, eg:

``` bash
$ helm install --name=keystone local/keystone --namespace=os-keystone
$ helm install --name=horizon local/horizon --namespace=os-horizon
$ helm install --name=glance local/glance --namespace=os-glance
$ helm install --name=nova local/nova --namespace=os-nova
$ helm install --name=neutron local/neutron --namespace=os-neutron
```

heat/templates/deployment/api/bin/_api.sh.tpl
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
```
DTadrzak: Instead of mounting script with only command inside execute it directly in api template.
```
command:
  - heat-api
  - --config-dir /etc/heat/conf
```

heat/templates/deployment/api/api.sh.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: heat-api-sh
+data:
+  start.sh: |+
```
DTadrzak: Unnecessary.

heat/templates/deployment/api/api.yaml
```diff
@@ -0,0 +1,76 @@
...
+        ]'
+    spec:
+      nodeSelector:
+        {{ .Values.labels.node_selector_key }}: {{ .Values.labels.node_selector_value }}
+      containers:
+        - name: heat-api
+          image: {{ .Values.images.api }}
+          imagePullPolicy: {{ .Values.images.pull_policy }}
+          command:
+          - bash
```
DTadrzak: change it to heat-api --config-dir /etc/heat/conf

heat/templates/deployment/api/api.yaml
```diff
@@ -0,0 +1,76 @@
...
+          - /tmp/start.sh
+          ports:
+            - containerPort: {{ .Values.service.api.port }}
+          readinessProbe:
+            tcpSocket:
+              port: {{ .Values.service.api.port }}
+          volumeMounts:
+            - name: heat-api-sh
```
DTadrzak: remove it

heat/templates/deployment/api/api.yaml
```diff
@@ -0,0 +1,76 @@
...
+              mountPath: /etc/heat/api-paste.ini
+              subPath: api-paste.ini
+              readOnly: true
+            - name: heat-conf-api
+              mountPath: /etc/heat/conf/heat-api.conf
+              subPath: heat-api.conf
+              readOnly: true
+{{ include "heat_config_volume_mounts" . | indent 12 }}
+      volumes:
+        - name: heat-api-sh
```
DTadrzak: remove it

heat/templates/deployment/cfn/bin/_cfn.sh.tpl
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
```
DTadrzak: Same here I think that we shouldn't create scripts that consists of only one command :)

heat/templates/deployment/cloudwatch/bin/_cloudwatch.sh.tpl
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
```
DTadrzak: ditto

heat/templates/jobs/db/init/bin/_db-init.sh.tpl
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+set -ex
+export HOME=/tmp
+
+ansible localhost -vvv -m mysql_db -a "login_host='{{ .Values.database.address }}' login_port='{{ .Values.database.port }}' login_user='{{ .Values.database.root_user }}' login_password='{{ .Values.database.root_password }}' name='{{ .Values.database.heat_database_name }}'"
```
DTadrzak: +1 let's make it more readable.

heat/templates/jobs/db/sync/bin/_db-sync.sh.tpl
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
```
DTadrzak: Same here

heat/templates/statefulset/engine/bin/_engine.sh.tpl
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
```
DTadrzak: ditto

heat/templates/deployment/api/bin/_api.sh.tpl
intlabs: @DTadrzak I believe that though this requires refactoring, that it should still be a script via a configmap, as this allows operators to easily customize as desired.

@alanmeadows could you confirm?

heat/templates/deployment/api/api.sh.yaml
intlabs: See above

heat/templates/deployment/api/api.yaml
intlabs: See above

heat/templates/deployment/api/api.yaml
intlabs: See above

heat/templates/deployment/api/api.yaml
intlabs: See above

heat/templates/jobs/keystone/endpoints/ks-endpoints.yaml
alanmeadows: Completely agree.  Clearly not everything is in place to do that today but it is the basic building block we'll need to support hyperscale.  With this support in place it becomes simple to (as an example) provide a dedicated rabbit cluster to ceilometer, or a database cluster per service.  These are really distinguishing features in my mind about what we're building.

It will eventually get there but we do need to keep it in mind when we lock in on the endpoint stuff.

heat/templates/deployment/api/bin/_api.sh.tpl
alanmeadows: I definitely get where you're trying to go but I would question whether it really provides significant value in its current form to the operator because to customize it they can't do it through helm, they must edit the template. Wouldn't the same be true of @DTadrzak's approach with the native command (in this editing the manifest itself) except in the latter case you get the benefit of less sprawl.

intlabs:
The next commit will refactor this work against the current master

heat/templates/bin/_ks-domain-user.sh.tpl
```diff
@@ -0,0 +1,55 @@
+#!/bin/bash
```
alanmeadows: I would like to see this script moved to common, understanding it has unique handling of keystone, but no reason we couldn't make the common script supportive of that.  Not necessary in this PR.

heat/templates/etc/_heat.conf.tpl
```diff
@@ -0,0 +1,84 @@
...
+debug = {{ .Values.misc.debug }}
+use_syslog = False
+use_stderr = True
+
+deferred_auth_method = "trusts"
+
+enable_stack_adopt = "True"
+enable_stack_abandon = "True"
+
+heat_metadata_server_url = {{ .Values.service.cfn.proto }}://{{ .Values.service.cfn.name }}:{{ .Values.service.cfn.port }}
```
alanmeadows: Seems like even when constructing "self-referencing" urls, we should be able to use the same routines we use to feed keystone for catalog insertion.  Otherwise, two ways of constructing endpoints and the goal should be a unified mechanism globally long-term.  Fine if this is a different PR.

heat/templates/job-ks-endpoints.yaml.yaml
```diff
@@ -0,0 +1,65 @@
...
+          volumeMounts:
+            - name: ks-endpoints-sh
+              mountPath: /tmp/ks-endpoints.sh
+              subPath: ks-endpoints.sh
+              readOnly: true
+          env:
+{{- with $env := dict "ksUserSecret" $ksAdminSecret }}
+{{- include "env_ks_openrc_tpl" $env | indent 12 }}
+{{- end }}
+            - name: OS_SVC_ENDPOINT
```
alanmeadows: There are complex namespace issues at play, but the end-goal I have is wrapping all common OS* vars like this into a call for a common template we feed input into, which can include name, type, and endpoint below.  This is much improved however over the previous version.  We can do this in a separate PR though.

heat/templates/job-ks-service.yaml
```diff
@@ -0,0 +1,59 @@
+{{- $envAll := . }}
+{{- $ksAdminSecret := .Values.keystone.admin_secret | default "heat-env-keystone-admin" }}
```
alanmeadows: I'm not a huge fan of default unless we need it to help construct something dynamic, involving the release namespace for instance. I like the values.yaml to be the engine for that otherwise as an operator you are popping open various manifest trying to figure out what value came from where.

heat/templates/job-ks-service.yaml
```diff
@@ -0,0 +1,59 @@
...
+          volumeMounts:
+            - name: ks-service-sh
+              mountPath: /tmp/ks-service.sh
+              subPath: ks-service.sh
+              readOnly: true
+          env:
+{{- with $env := dict "ksUserSecret" $ksAdminSecret }}
+{{- include "env_ks_openrc_tpl" $env | indent 12 }}
+{{- end }}
+            - name: OS_SERVICE_NAME
```
alanmeadows: Same comment as above.  I am trying to avoid the mix of stated envs (L52) vs dynamc ones (L48)  formed with include unless absolutely required.

heat/templates/job-ks-user.yaml
```diff
@@ -0,0 +1,124 @@
+{{- $ksAdminSecret := .Values.keystone.admin_secret | default "heat-env-keystone-admin" }}
```
alanmeadows: Same default comment as above.

heat/templates/job-ks-user.yaml
```diff
@@ -0,0 +1,124 @@
...
+          env:
+{{- with $env := dict "ksUserSecret" $ksAdminSecret }}
+{{- include "env_ks_openrc_tpl" $env | indent 12 }}
+{{- end }}
+            - name: SERVICE_OS_SERVICE_NAME
+              value: "heat"
+{{- with $env := dict "ksUserSecret" $ksUserSecret }}
+{{- include "env_ks_user_create_openrc_tpl" $env | indent 12 }}
+{{- end }}
+            - name: SERVICE_OS_ROLE
```
alanmeadows: Same as the mix comment above.

heat/templates/job-ks-user.yaml
```diff
@@ -0,0 +1,124 @@
...
+          volumeMounts:
+            - name: ks-user-sh
+              mountPath: /tmp/ks-domain-user.sh
+              subPath: ks-domain-user.sh
+              readOnly: true
+          env:
+{{- with $env := dict "ksUserSecret" $ksAdminSecret }}
+{{- include "env_ks_openrc_tpl" $env | indent 12 }}
+{{- end }}
+            - name: SERVICE_OS_SERVICE_NAME
```
alanmeadows: Again.

heat/values.yaml
```diff
@@ -0,0 +1,197 @@
+# Default values for keystone.
+# This is a YAML-formatted file.
+# Declare name/value pairs to be passed into your templates.
+# name: value
+
+replicas: 1
```
alanmeadows: Don't you need to break this out for the various api services?

heat/templates/etc/_heat.conf.tpl
```diff
@@ -0,0 +1,84 @@
...
+
+[trustee]
+auth_type = "password"
+auth_section = "trustee_keystone"
+
+[trustee_keystone]
+signing_dir = "/var/cache/heat"
+memcached_servers = "{{ .Values.memcached.host }}:{{ .Values.memcached.port }}"
+auth_version = v3
+auth_uri = {{ .Values.keystone.auth_uri }}
```
alanmeadows: Right or wrong for the moment, the other charts are using {{ include "endpoint_keystone_internal . }} here

heat/templates/etc/_heat.conf.tpl
```diff
@@ -0,0 +1,84 @@
...
+auth_url = {{ .Values.keystone.auth_url }}
```
alanmeadows: Right or wrong for the moment, the other charts are using {{ include "endpoint_keystone_admin . }} here

heat/values.yaml
intlabs: good catch, I'm sure I did but apparently not :)

heat/templates/job-ks-service.yaml
intlabs: Agreed, this also is in line with the draft Helm Conventions document that @so0k has been working on:
 * https://gist.github.com/so0k/f927a4b60003cedd101a0911757c605a

heat/templates/etc/_heat.conf.tpl
intlabs: I'll tackle this as part of a separate PR - as the URLs used here are not quite the same as the keystone registered endpoints. Though the endpoint function should still be used to do this.

heat/templates/etc/_heat.conf.tpl
intlabs: done

heat/templates/etc/_heat.conf.tpl
intlabs: done

heat/templates/job-ks-endpoints.yaml.yaml
intlabs: Again I will get this resolved in a sperate PR once I have worked out how to best deal with the namespacing issues.

heat/templates/job-ks-service.yaml
intlabs: As above

heat/templates/bin/_ks-domain-user.sh.tpl
intlabs: Moved

PR #79, opened 2017-01-03
adding an online github-based helm repository

v1k0d3n:
we'll have to maintain this with each pull request, but this should allow outside folks to use the charts simply by calling the gh-pages branch.

---
This change is [Reviewable](https://reviewable.io/reviews/att-comdev/openstack-helm/79)


alanmeadows:
Updating it for each pull request is likely not going to be sustainable, what I think you want to do is whenever there is a major release (e.g we've tagged master for 0.1.0 or beyond) and that version is fully tested as a complete stack, as part of that process we release to here as well.  Otherwise we can't guarantee each pull request is providing a valid chart with our lack of gate testing.  Even when we have basic gate testing it will be some time before we can guarantee full stack functionality.  A tagged release would be the time to be doing that full testing and building these.

v1k0d3n:
yeah, i agree...especially after thinking about it. i intend on doing it with each tagged release only. see my updated comments on the main [README.md](https://github.com/att-comdev/openstack-helm/blob/master/README.md#installation-and-development).

v1k0d3n:
all things considered, i do think we could update the gh-pages branch using CI with each merged PR. what are your thoughts about that?

intlabs:
@v1k0d3n: From my perspective an update on each merged PR would be super sweet.

Once there is some stability to the project, I would have thought that ideally there would be two streams/repos - one following `releases` as marked by git repo tags (and possibly manually managed), and another bleeding edge repo that followed `master`.

v1k0d3n:
@intlabs yup, i think doing releases in branches, and then tagging appropriately would be best for the project going forward. once we have an initial release (after our Jan 10th date), we can start following more closely with Openstack releases and tagged updates.

PR #83, opened 2017-01-03
Stateful ceph

alanmeadows:
Refactor ceph to allow monitors to leverage a stateful set.
    
The IPs of ceph monitors cannot (or should not) change.
    
This refactor allows the ceph monitors to act as statefulsets.
It also persists their on disk data to nodeDirs, to allow cluster
wide restarts (in parallel), where previously this would lose
data.  This is accompanied by some docker images changes that
ensure that auth for the OSDs is restored/reinserted as auth
data is somehow not persisted to disk at this time.

---
This change is [Reviewable](https://reviewable.io/reviews/att-comdev/openstack-helm/83)


v1k0d3n:
Done.

---

Reviewed 25 of 25 files at r1.
Review status: all files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/att-comdev/openstack-helm/83#-:-K_a8_5OiBzs0Gf3IEwZ:b-896fix)*

v1k0d3n:
much cleaner! great PR alan.

we should really start using the values that you've included in this chart in some of the others:
i.e. fqdn, removal of hard api endpoints, default passwords that are the same across charts, etc. this is a really good ceph chart. thanks for your help too @intlabs!

keystone/templates/etc/_keystone.conf.tpl
```diff
@@ -9,11 +9,11 @@ connection = mysql+pymysql://{{ .Values.database.keystone_user }}:{{ .Values.dat
 max_retries = -1
 
 [memcache]
-servers = {{ include "memcached_host" . }}
+servers = {{ include "memcached_host" . }}:11211
```
intlabs: I've been thinking about how to handle memcached, regarding scaling and security. Though for now, I think this and the way we deploy memcached is fine, after chatting with a few folks think we should possibly move memcached to a stateful set and use a headless service along with an init container to build the config dynamically on pod start. Just food for thought, I'll write this up better as a proposal when I get time for consideration.

alanmeadows: For sure - our rabbit and memcache charts are what we call "Toy" charts for now.

They lack any kind of resiliency or scaling and are just building blocks to stand things up.

We discussed with the intel folks that we need a concerted plan around resiliency and scaling for memcache and rabbit and we should all come together on what that should be.

Even the host and endpoint work still needs some thought but I want to set up something--even if it means I need to refactor later.  Because of our speed, none of this is getting the thought it should yet but I wanted to do something, even if it generates negative feedback.

PR #87, opened 2017-01-04
maas - adding support for postgres volume mount

larryrensing:
initial commit, currently using hostDir to mimic pvc

---
This change is [Reviewable](https://reviewable.io/reviews/att-comdev/openstack-helm/87)


v1k0d3n:
Review status: 0 of 14 files reviewed at latest revision, 6 unresolved discussions, some commit checks failed.

---

*[Makefile, line 9 at r1](https://reviewable.io:443/reviews/att-comdev/openstack-helm/87#-K_dnenocCmk2XhhFNX-:-K_dnenocCmk2XhhFNX0:by9i61c) ([raw file](https://github.com/att-comdev/openstack-helm/blob/aff7ef7d029b8db44b1db7ecd11ce27bc036fc28/Makefile#L9)):*
> all: common bootstrap mariadb rabbitmq maas memcached keystone openstack

Can you rebase your changes so it avoids merge conflicts with what's been pushed since you initially forked out air-helm? from what i can tell, this file is the main culprit.

---

*[maas/requirements.lock, line 2 at r1](https://reviewable.io:443/reviews/att-comdev/openstack-helm/87#-K_dnYd0-XqZugezeWRt:-K_dnYd0-XqZugezeWRu:blqd1j7) ([raw file](https://github.com/att-comdev/openstack-helm/blob/aff7ef7d029b8db44b1db7ecd11ce27bc036fc28/maas/requirements.lock#L2)):*
> dependencies:
> - name: common
>   repository: http://localhost:8879/charts
>   version: 0.1.0
> digest: sha256:c6a7e430c900036912fe3fdc14213e9280c5da0b6607ce4dcf6dc95535d114fc
> generated: 2017-01-03T19:36:04.854911655Z

put a *.lock in your .gitignore. this change was added after your initial fork-out for maas.

---

*[maas/values.yaml, line 6 at r1](https://reviewable.io:443/reviews/att-comdev/openstack-helm/87#-K_dn9g38wvU4aY3GOmU:-K_dn9g38wvU4aY3GOmV:b-efy0at) ([raw file](https://github.com/att-comdev/openstack-helm/blob/aff7ef7d029b8db44b1db7ecd11ce27bc036fc28/maas/values.yaml#L6)):*
>   maas_region: quay.io/larryrensing/maas-region:1.1

submit changes to the current container in our attcomdev/dockerfiles repo, and tag them according to the version of MaaS (so other community members can quickly identify with versions that make sense from a MaaS POV).

also, i think i changed the container tags to reflect the versions of MaaS used in the image. you may want to try with this update.

---

*[maas/templates/service.yaml, line 11 at r1](https://reviewable.io:443/reviews/att-comdev/openstack-helm/87#-K_dn4cOUrG841zTHv-l:-K_dn4cOUrG841zTHv-m:b-xbtwhk) ([raw file](https://github.com/att-comdev/openstack-helm/blob/aff7ef7d029b8db44b1db7ecd11ce27bc036fc28/maas/templates/service.yaml#L11)):*
> spec:
>   type: NodePort
>   ports:
>   - port: {{ .Values.network.port.service_gui }}
>     targetPort: {{ .Values.network.port.service_gui_target }}

fine for now, but we may want to make this configurable for folks who are using ingress (which we will be doing soon as well. doesn't need to be changed now, but we need to track it as an issue and get it scheduled along with any ingress work we do.

---

*[maas/templates/etc/_region-dns-config.tpl, line 3 at r1](https://reviewable.io:443/reviews/att-comdev/openstack-helm/87#-K_dn-gMMpqG2ZmRvqzb:-K_dn-gMMpqG2ZmRvqzc:bwepxf4) ([raw file](https://github.com/att-comdev/openstack-helm/blob/aff7ef7d029b8db44b1db7ecd11ce27bc036fc28/maas/templates/etc/_region-dns-config.tpl#L3)):*
> listen-on-v6 { any; };

we may want to turn off any IPv6 for now until we truly implement it in the future. just to keep things clean (network-wise).

---

*[maas/templates/var/_maas-region-controller.postinst.tpl, line 47 at r1](https://reviewable.io:443/reviews/att-comdev/openstack-helm/87#-K_dmvG2Nv6NZ7ovBPlE:-K_dmvG2Nv6NZ7ovBPlF:b-z6yoko) ([raw file](https://github.com/att-comdev/openstack-helm/blob/aff7ef7d029b8db44b1db7ecd11ce27bc036fc28/maas/templates/var/_maas-region-controller.postinst.tpl#L47)):*
> get_default_route_ip6() {
<details><summary><i alt="summary">Quoted 8 lines of code…</i></summary>
>     while read Src SrcPref Dest DestPref Gateway Metric RefCnt Use Flags Iface
>     do
>         [ "$SrcPref" = 00 ] && [ "$Iface" != lo ] && break
>     done < /proc/net/ipv6_route
>     if [ -n "$Iface" ]; then
>         LC_ALL=C /sbin/ip -6 addr list dev "$Iface" scope global permanent |
>             sed -n '/ inet6 /s/.*inet6 \([0-9a-fA-F:]*\).*/[\1]/p' | head -1
>     fi
</details>
> }

again, see IPv6 notes.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/att-comdev/openstack-helm/87)*

v1k0d3n:
@larryrensing, for now, just work on the rebase and I will merge. we need this chart for testing 0.1.0 release, and it's fine "as is". then submit another PR for the notes I made above.

larryrensing:
rebased and ignoring .lock files in gitignore -- it should be good now

also uploaded new maas region docker image to quay: quay.io/attcomdev/maas-region:2.1.2-1

maas/values.yaml
```diff
@@ -3,8 +3,8 @@
 # Declare variables to be passed into your templates.
 
 images:
-  maas_region: quay.io/attcomdev/maas-region:1.0.1
-  maas_rack: quay.io/attcomdev/maas-rack:1.0.1
+  maas_region: quay.io/attcomdev/maas-region:2.1.2-1
+  maas_rack: quay.io/attcomdev/maas-rack:2.1.2
```
v1k0d3n: for confusion purposes, create a 2.1.2-1 for rack too.

v1k0d3n:
going to merge. if you could tag the container that would be good. also, create an issue to refactor so usernames and passwords are in `Values.yaml`. that's really important for the chart going forward.

maas/templates/_helpers.tpl
```diff
@@ -0,0 +1,3 @@
+{{- define "joinListWithColon" -}}
```
alanmeadows: This is in common now.  No need to define it here.  _helpers.tpl can be removed.

maas/templates/deploy-region.yaml
```diff
@@ -1,12 +1,55 @@
...
+        pod.beta.kubernetes.io/init-containers: '[
+          {
+            "name": "init",
+            "image": "{{ .Values.images.maas_region }}",
+            "imagePullPolicy": "Always",
+            "command": [
+                    "/bin/bash", "-c"
+            ],
+            "args": [
+                    "chmod +x /tmp/start.sh; /tmp/start.sh"
```
alanmeadows: The chmod should not be necessary here if you use ['bash', '/tmp/start.sh']

maas/templates/service.yaml
```diff
@@ -1,10 +1,11 @@
 apiVersion: v1
 kind: Service
 metadata:
-  name: maas-region-ui
+  name: {{ .Values.service_name }}
   labels:
     app: maas-region-ui
 spec:
+  type: NodePort
```
alanmeadows: This may of helped with testing, but no other chart specifically leverages a NodePort (could be conflicts if they all did).

We will do a cross-chart ingress approach once we've tackled that.

maas/templates/var/_maas-region-controller.postinst.tpl
```diff
@@ -0,0 +1,149 @@
...
+    # if default-maas-url has not been preseeded.  Prefer ipv4 addresses if
+    # present, and use "localhost" only if there is no default route in either
+    # address family.
+    db_get maas/default-maas-url
+    ipaddr="$RET"
+    if [ -z "$ipaddr" ]; then
+        #ipaddr=$(get_default_route_ip4)
+        ipaddr="maas-region-ui.{{ .Release.Namespace }}"
+    fi
+    if [ -z "$ipaddr" ]; then
```
alanmeadows: You can remove these two other conditionals as $ipaddr at this point would never be null.

maas/templates/deploy-region.yaml
```diff
@@ -1,12 +1,55 @@
...
+                },
+                {
+                    "name": "maasregionpostinst",
+                    "mountPath": "/var/lib/dpkg/info/maas-region-controller.postinst",
+                    "subPath": "maas-region-controller.postinst"
+                }
+            ]
+          }
+        ]'
     spec:
```
alanmeadows: I would advise a readinessProbe for the region controller to validate port 80 is responsive.

maas/templates/configmap-var.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: v1
```
alanmeadows: I think the configmap-etc and configmap-bin names may of been misleading.

Rather then path prefixes, its really "scripts" vs "config" or "bin" vs "etc" so configmap-var is not wholly necessary.

larryrensing:
will create another PR with @alanmeadows suggested changes

PR #45, opened 2016-12-14
[WIP] Nova support

DTadrzak:
>>WIP do not merge<<

Adding: nova-api, nova-scheduler, etc services
Adding: nova-compute and nova-libvirt as a deployment
Adding: support for kubernetes drain commond to nova-compute

Signed-off-by: DTadrzak <daniel.tadrzak@intel.com>

PR #91, opened 2017-01-04
OpenStack Nova Helm Chart Initial Commit

alanmeadows:
This is an initial commit of a functional nova chart.  It has been
tested with the neutron chart using a flat network configuration.

---
This change is [Reviewable](https://reviewable.io/reviews/att-comdev/openstack-helm/91)


alanmeadows:
The bulk of the effort was accomplished by @DTadrzak's existing Nova WIP closed in https://github.com/att-comdev/openstack-helm/pull/45

nova/templates/etc/_nova.conf.tpl
```diff
@@ -0,0 +1,108 @@
...
+
+[neutron]
+url = {{ include "endpoint_neutron_api_internal" . }}
+
+metadata_proxy_shared_secret = {{ .Values.neutron.metadata_secret }}
+service_metadata_proxy = True
+
+auth_url = {{ include "endpoint_keystone_admin" . }}
+auth_type = password
+project_domain_name = default
```
DTadrzak: In future we should make it configurable.

nova/templates/job-db-sync.yaml
```diff
@@ -0,0 +1,56 @@
...
+              }
+            ]
+          }
+        ]'
+    spec:
+      restartPolicy: OnFailure
+      containers:
+        - name: nova-db-sync
+          image: {{ .Values.image.db_sync }}
+          imagePullPolicy: Always
```
DTadrzak: s/Always/{{ .Values.image.pull_policy }}/

nova/templates/job-init.yaml
```diff
@@ -0,0 +1,50 @@
...
+              }
+            ]
+          }
+        ]'
+    spec:
+      restartPolicy: OnFailure
+      containers:
+        - name: nova-init
+          image: {{ .Values.image.init }}
+          imagePullPolicy: Always
```
DTadrzak: Same here

nova/templates/job-post.yaml
```diff
@@ -0,0 +1,59 @@
...
+              }
+            ]
+          }
+        ]'
+    spec:
+      restartPolicy: OnFailure
+      containers:
+        - name: nova-post
+          image: {{ .Values.image.post }}
+          imagePullPolicy: Always
```
DTadrzak: s/Always/{{ .Values.image.pull_policy }}

nova/templates/service-api.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: v1
```
DTadrzak: @intlabs wrote in PR #45 "Yoy may want to split these into two services, the nova-metadata service requires some special handling with many neutron configurations."

nova/values.yaml
```diff
@@ -0,0 +1,213 @@
...
+  control_node_selector_key: openstack-control-plane
+  control_node_selector_value: enabled
+  compute_node_selector_key: openstack-compute-node
+  compute_node_selector_value: enabled
+
+control_replicas: 1
+compute_replicas: 1
+
+image:
+  init: quay.io/stackanetes/stackanetes-kolla-toolbox:barcelona
```
DTadrzak: s/barcelona/newton/g

nova/values.yaml
```diff
@@ -0,0 +1,213 @@
...
+  db_sync: quay.io/stackanetes/stackanetes-nova-api:barcelona
+  api: quay.io/stackanetes/stackanetes-nova-api:barcelona
+  conductor: quay.io/stackanetes/stackanetes-nova-conductor:barcelona
+  scheduler: quay.io/stackanetes/stackanetes-nova-scheduler:barcelona
+  novncproxy: quay.io/stackanetes/stackanetes-nova-novncproxy:barcelona
+  consoleauth: quay.io/stackanetes/stackanetes-nova-consoleauth:barcelona
+  compute: quay.io/stackanetes/stackanetes-nova-compute:barcelona
+  libvirt: quay.io/stackanetes/stackanetes-nova-libvirt:barcelona
+  post: quay.io/stackanetes/stackanetes-kolla-toolbox:barcelona
+  entrypoint: quay.io/stackanetes/kubernetes-entrypoint:v0.1.0
```
DTadrzak: I think that entrypoint and pull_policy should be moved to global

nova/values.yaml
```diff
@@ -0,0 +1,213 @@
...
+    novncproxy: 6080
+
+nova:
+  default:
+    debug: false
+    osapi_workers: 8
+    metadata_workers: 8
+    conductor_workers: 8
+    force_config_drive: True
+  drain_timeout: 60
```
DTadrzak: We don't have drain support right now :)

nova/values.yaml
```diff
@@ -0,0 +1,213 @@
...
+  secret_uuid: ""
+
+neutron:
+  metadata_secret: "password"
+
+memcached:
+  address: "memcached:11211"
+
+dependencies:
+  api:
```
intlabs: Quite a lot here, I'm not very comfortable bout the amount of possibly superfluous checking we are doing of the environment, as it will make scaling across clusters and recovery from failure potentially much harder to manage, some food for thought:

(also db_sync is in twice)
```
dependencies:
  api:
    jobs:
    - nova-db-sync
# Note: I rather cheekily put the post here, this may not be appropriate at the moment.
    - nova-post
    service:
    - mariadb
  db_sync:
    jobs:
    - nova-init
    service:
    - mariadb
# Actually not very comfortable about the whole 'post' job
# It's good practice to set this stuff up before the services come online
# I think that the approach I took for heat and cinder (though a WIP,
# and needing refactoring) Is much cleaner.
  post:
    service:
    - keystone-api
  init:
    jobs:
    - mariadb-seed
    service:
    - mariadb
  compute:
    service:
    - nova-api
    daemonset:
    - nova-libvirt
# Need to check but does libvirt deamon actually have any deps?
  libvirt:
    service:
    - keystone-api
    - nova-api
  consoleauth:
    jobs:
    - nova-db-sync
    service:
    - mariadb
  scheduler:
    jobs:
    - nova-db-sync
    service:
    - mariadb
  conductor:
    jobs:
    - nova-db-sync
    service:
    - mariadb
```

The about is just an example - I've not checked it - but shows the potential for trimming. In essence why check that the (for example) the `keystone-init` job has run when all we need is the `keystone-api` to be up?

nova/values.yaml
DTadrzak: Sure, all those images are built by Kolla (debian based). There is only 1 extra layer.

nova/values.yaml
alanmeadows: I've incorporated the majority of this, some of it requires more investigation.  This direction though sets things up for more flexibility.  For instance, just as an example, the more we can remove things like "mariadb" or label it as "infra-db" (similar to the keystone->identity endpoint discussion) allows operators to make database swaps.

nova/values.yaml
alanmeadows: Removed.

nova/values.yaml
alanmeadows: Still leaving this here as I still assert that its _possible_ an operator might want a different pull policy on a per chart basis.  Think we need to discuss more.

nova/values.yaml
alanmeadows: We will once we have pipelines in place to build them (getting close).

nova/templates/service-api.yaml
alanmeadows: Split out metadata and API.

nova/templates/job-post.yaml
alanmeadows: Fixed.

nova/templates/job-init.yaml
alanmeadows: Fixed.

nova/templates/job-db-sync.yaml
alanmeadows: Fixed.

nova/templates/etc/_nova.conf.tpl
alanmeadows: Fixed

nova/templates/deployment-api.yaml
```diff
@@ -0,0 +1,70 @@
...
+          }
+        ]'
+    spec:
+      nodeSelector:
+        {{ .Values.labels.control_node_selector_key }}: {{ .Values.labels.control_node_selector_value }}
+      containers:
+        - name: nova-api
+          image: {{ .Values.image.api }}
+          # https://bugs.launchpad.net/kolla-mesos/+bug/1546007
+          securityContext:
```
alanmeadows: This has been implemented.

nova/templates/daemonset-compute.yaml
```diff
@@ -0,0 +1,138 @@
...
+          image: {{ .Values.image.compute }}
+          imagePullPolicy: Always
+          securityContext:
+            privileged: true
+          command:
+            - nova-compute
+            - --config-file
+            - /etc/nova/nova.conf
+          env:
+            - name: POD_NAME
```
alanmeadows: Removed.

nova/templates/bin/_start-osapi.sh.tpl
```diff
@@ -0,0 +1,21 @@
...
+true
+EOF
+
+# make it executable and copy it over whatever iptables may be underneath in this image
+chmod +x /tmp/iptables
+cp -p /tmp/iptables /sbin/iptables
+cp -p /tmp/iptables /sbin/iptables-restore
+cp -p /tmp/iptables /sbin/iptables-save
+
+exec nova-api--config-file /etc/nova/nova.conf
```
intlabs: Missing space between `nova-api` and `--config-file`.

alanmeadows: Fixed

DTadrzak:
@intlabs We can also apply that fix directly in image. AFAIK image on quay.io is already fixed.

intlabs:
@DTadrzak we can, and that's what I do with my old ones as well - but to ensure we are image agnostic I think the solution Alan has made is better.

Issue #89, opened 2017-01-04
Ensure init-containers are consistently used throughout every chart for dependency checks


alanmeadows:
We need to ensure that every chart consistently leverages the init-container approach.

An example of where we are doing it correctly:

https://github.com/att-comdev/openstack-helm/blob/master/keystone/templates/deployment.yaml#L11-L40

An example of something that needs conversion:

https://github.com/att-comdev/openstack-helm/blob/master/neutron/templates/deployment-server.yaml#L14-L34

alanmeadows:
Also relates to https://github.com/att-comdev/openstack-helm/issues/69

alanmeadows:
As part of this effort, we also need to remove unnecessary environment variables, e.g. in https://github.com/att-comdev/openstack-helm/pull/84/files/4d1b5320f7864462657fe93df498b3e3c4db6bca#diff-54d43c699714abd5ec5ca3feb92891ae

Ciello89:
@alanmeadows if you're not currently working on this, I can take it over as part of the refactoring I'm doing since yesterday on the initial chart for neutron.

alanmeadows:
@Ciello89 It's all yours, this will be a great consistency/cleanup win.

intlabs:
@alanmeadows I've done this as part of https://github.com/att-comdev/openstack-helm/pull/138

ss7pro:
Folks, as init-containers are going to be moved out of beta in K8S 1.6 we should ensure that our charts would be able to handle both of them. We should go even further with init-container template to included event the calling statement in the common package (so we can have  have condition on k8s version in a single place).

Issue #88, opened 2017-01-04
Investigate replacement for stackanetes-entrypoint


alanmeadows:
The stackanetes-entrypoint image consistently leveraged runs into issues with DAEMONSET dependencies, and has trouble with config file dependencies.  These should be documented in this issue subsequently but I do not have readily available access to tracebacks at the moment.

We either need to:

1) Investigate whether since its creation there is a now "standard" init container for dependency checking that is being widely used

2) If we continue to use it, ensure we are including it in our CI/CD pipelines we will develop and be creating issues to have the issues we encountered resolved in the upstream repository: https://github.com/stackanetes/kubernetes-entrypoint

alanmeadows:
@DTadrzak @PiotrProkop 

Currently, when trying to use the daemonset dependency, I get this:

```
ubuntu@MT01DB04:~$ kubectl run -i --tty entrypoint-test --image=quay.io/stackanetes/kubernetes-entrypoint:v0.1.0 --restart=Never --env=NAMESPACE=default,DEPENDENCY_DAEMONSET=foobar,COMMAND="echo done" -- /kubernetes-entrypoint
Waiting for pod default/entrypoint-test to be running, status is Pending, pod ready: false
Waiting for pod default/entrypoint-test to be running, status is Pending, pod ready: false
Entrypoint Error: 2017/01/05 19:40:40 daemonset.go:25: Cannot initialize daemonset: Env POD_NAME not set
panic: value method daemonset.Daemonset.GetName called using nil *Daemonset pointer

goroutine 9 [running]:
panic(0x127b240, 0xc420427e90)
	/usr/local/go/src/runtime/panic.go:500 +0x1a1
github.com/stackanetes/kubernetes-entrypoint/dependencies/daemonset.(*Daemonset).GetName(0x0, 0x147b318, 0xc420427e80)
	<autogenerated>:2 +0xa4
github.com/stackanetes/kubernetes-entrypoint/entrypoint.Entrypoint.Resolve.func1(0xc420427e80, 0x1bd2f00, 0xc4203389c0, 0xc42001202a, 0x7, 0x1bc7b00, 0x0)
	/home/pprokop/gopath/src/github.com/stackanetes/kubernetes-entrypoint/entrypoint/entrypoint.go:61 +0x8b
created by github.com/stackanetes/kubernetes-entrypoint/entrypoint.Entrypoint.Resolve
[...]

PR #93, opened 2017-01-05
Resolve neutron feedback from PR#60

alanmeadows:
This commit addresses:

* Separating out stacked ovs daemonset into separate daemonsets.

* Fixes line ending issues.

* Enhances agents ovs pre-flight checks by using neutron-sanity-check since we can no longer rely on container dependencies and daemonset dependencies in entrypoint are not yet working.

---
This change is [Reviewable](https://reviewable.io/reviews/att-comdev/openstack-helm/93)


neutron/templates/daemonset-ovs-agent.yaml
```diff
@@ -0,0 +1,86 @@
...
+            - name: NAMESPACE
+              valueFrom:
+                fieldRef:
+                  fieldPath: metadata.namespace
+            - name: COMMAND
+              value: "bash /tmp/neutron-openvswitch-agent.sh"
+            - name: DEPENDENCY_JOBS
+              value: "{{  include "joinListWithColon" .Values.dependencies.ovs_agent.jobs }}"
+            - name: DEPENDENCY_SERVICE
+              value: "{{  include "joinListWithColon" .Values.dependencies.ovs_agent.service }}"
```
intlabs: We should move this (dependency checks) to an init-container - to both come in line with other services and provide flexibility in the images we can use.

neutron/templates/daemonset-ovs-db.yaml
```diff
@@ -0,0 +1,61 @@
...
+            - name: POD_NAME
+              valueFrom:
+                fieldRef:
+                  fieldPath: metadata.name
+            - name: NAMESPACE
+              valueFrom:
+                fieldRef:
+                  fieldPath: metadata.namespace
+            - name: COMMAND
+              value: "bash /tmp/openvswitch-db-server.sh"
```
intlabs: As above (move to init-container).

neutron/templates/daemonset-ovs-vswitchd.yaml
```diff
@@ -0,0 +1,67 @@
...
+            - name: POD_NAME
+              valueFrom:
+                fieldRef:
+                  fieldPath: metadata.name
+            - name: NAMESPACE
+              valueFrom:
+                fieldRef:
+                  fieldPath: metadata.namespace
+            - name: COMMAND
+              value: "bash /tmp/openvswitch-vswitchd.sh"
```
intlabs: As above (move to init container).

neutron/values.yaml
```diff
@@ -159,17 +159,14 @@ dependencies:
     - nova-api
     daemonset:
     - neutron-openvswitch
-  openvswitchagent:
+  ovs_agent:
```
intlabs: Think this whole section could do with reviewing (though possibly out of scope for this ps). Off the top of my head it could be changed to:

```
dependencies:  
  server:
    jobs:
    - neutron-db-sync
    service:
    - rabbitmq
    - mariadb
    - keystone-api
    - memcached
  dhcp:
    service:
    - neutron-server
    - rabbitmq
    - nova-api
    daemonset:
    - neutron-openvswitch
  metadata:
    service:
    - neutron-server
    - nova-api
    daemonset:
    - neutron-openvswitch
  ovs_agent:
    service:
    - keystone-api
    - rabbitmq
    - neutron-server
  l3:
    service:
    - neutron-server
    - rabbitmq
    - nova-api
    daemonset:
    - neutron-openvswitch    
  db_sync:
    jobs:
    - neutron-init
    service:
    - mariadb
  init:
    jobs:
    - mariadb-seed
    service:
    - mariadb
  post:
    jobs:
    - neutron-db-sync
    service:
    - keystone-api
```

neutron/templates/daemonset-ovs-vswitchd.yaml
intlabs: As above (change to init container)

neutron/templates/bin/_neutron-openvswitch-agent.sh.tpl
```diff
@@ -2,6 +2,16 @@
...
 chown neutron: /run/openvswitch/db.sock
 
+# ensure we can talk to openvswitch or bail early
+# this is until we can setup a proper dependency
+# on deaemonsets - note that a show is not sufficient
+# here, we need to communicate with both the db and vswitchd
+# which means we need to do a create action
+# 
+# see https://github.com/att-comdev/openstack-helm/issues/88
+timeout 3m neutron-sanity-check --config-file /etc/neutron/neutron.conf --ovsdb_native --nokeepalived_ipv6_support
```
intlabs: This is cool, I was not aware of this command.

neutron/templates/daemonset-ovs-agent.yaml
alanmeadows: Agreed, plan is to tackle that as part of #89

neutron/values.yaml
alanmeadows: Sure. Let me examine the diff and get back to you on this.

neutron/templates/bin/_neutron-openvswitch-agent.sh.tpl
alanmeadows: It is quite handy, even with the dependency stuff in place it makes sense at some level to take advantage of it as it can catch errors early and exit.  The neutron agent tends to mask errors such as failures working with br-int and just hang or skip foundational setup which is a terrible thing to do because it 'looks' like its working when it will never properly pass packets.

neutron/values.yaml
alanmeadows: This has been addressed.  Some of the jobs cannot be removed as I imagine you were collapsing them because of the daemonset dependency which would by its nature ensure they completed.  We are not yet leveraging the daemonset dependencies because of https://github.com/att-comdev/openstack-helm/issues/88 but otherwise, all other suggestions applied.

neutron/templates/daemonset-ovs-db.yaml
```diff
@@ -0,0 +1,61 @@
...
+          emptyDir: {}
+        - name: resolvconf
+          configMap:
+            name: neutron-etc
+        - name: libmodules
+          hostPath:
+            path: /lib/modules
+        - name: run
+          hostPath:
+            path: /run
```
DTadrzak: new line :)

PR #94, opened 2017-01-08
Add a Codacy badge to README.md

codacy-badger:
This change is [Reviewable](https://reviewable.io/reviews/att-comdev/openstack-helm/94)


v1k0d3n:
i don't like this PR and we're not going to use Codacy that I can see.
created in error; looking at other options.

Issue #69, opened 2016-12-27
feat: Init containers should be moved to a template in common/ utils 


ss7pro:
This would be very helpful when k8s would move init containers out of annotations.

intlabs:
I think this is completed with https://github.com/att-comdev/openstack-helm/pull/137

v1k0d3n:
@ss7pro; I believe this work is done. can you verify that this targets what you were asking for?

PR #104, opened 2017-01-11
Init-container templates

DTadrzak:
PR relates to #69 and #89 

Adding: templates for init-containers. Unfortunately I need to break it for
two templates, because i don't how to pass two variables to template.

PR also adjust layout of configuration files in Nova.

---
This change is [Reviewable](https://reviewable.io/reviews/att-comdev/openstack-helm/104)


cinder/templates/deployment-api.yaml
```diff
@@ -19,31 +19,8 @@ spec:
...
-                "value": "{{  include "joinListWithColon" .Values.dependencies.api.jobs }}"
-              },
-              {
-                "name": "COMMAND",
-                "value": "echo done"
-              }
-            ]
-          }
-        ]'
+      {{  include "init-containers-header" . }}
```
alanmeadows: To collapse this you could potentially follow the pattern of "template":

https://github.com/att-comdev/openstack-helm/blob/master/common/templates/_funcs.tpl#L5-L7

Using the tuple method to pass in the arguments referenced with index:

https://github.com/att-comdev/openstack-helm/blob/92512d66bb45d078775c92d324dbedb85a686536/neutron/templates/configmap-etc.yaml#L7

I don't know of a better way - I have expressed to @technosophos on #helm that being able to provide input to template "blocks" needs to be more intuitive.

common/templates/_funcs.tpl
```diff
@@ -21,3 +21,42 @@
 {{- $wtf := $context.Template.Name | replace $last $name -}}
 {{- include $wtf $context | sha256sum | quote -}}
 {{- end -}}
+
+{{- define "init-containers-header"}}
```
alanmeadows: I worry this precludes people from adding "chart specific" init containers to the charts - if you shifted the template to start at ```{ "name" ....}``` they could definite additional init containers that are not common?

alanmeadows: We may also want to be specific with this define name, its specifically a dependency check init container.

intlabs:
Hey, this looks like a good start, but I think needs a few changes:
- Should be reworked to be a single template (you may find the stuff I did for endpoint registration useful here)
- The template should only contain the init container itself, allowing us to add additional init containers to a pod as required. This can be a bit tricky with the current json implementation, but I think having to restrict the init container to the last element is an acceptable compromise for now.
- Obviously, I appreciate that this is a WIP, but the development helper scripts should be removed.

common/templates/_funcs.tpl
```diff
@@ -22,37 +22,34 @@
...
-      }
-    ]
-  }
-]'
+{{- define "dep-check-init-cont" -}}
+{{- $envALL := index . 0 -}}
+{{- $deps := index . 1 -}}
+{
+  "name": "init",
+  "image": {{ $envALL.Values.images.dep_check | default "quay.io/stackanetes/kubernetes-entrypoint:v0.1.0" | quote }},
```
intlabs: We shouldn't have an image declared as a default.

cinder/templates/deployment-api.yaml
```diff
@@ -19,8 +21,9 @@ spec:
       annotations:
         configmap-bin-hash: {{ tuple "configmap-bin.yaml" . | include "hash" }}
         configmap-etc-hash: {{ tuple "configmap-etc.yaml" . | include "hash" }}
-{{  include "dep-check-init-cont-header" . | indent 8 }}
-{{  include "dep-check-init-cont-footer" .Values.dependencies.api | indent 8 }}
+        pod.beta.kubernetes.io/init-containers: '[
+{{ tuple $envAll $dependecies | include "dep-check-init-cont" | indent 10 }}
```
intlabs: Typo, this affects all instances:
s/dependecies/dependencies

common/templates/_funcs.tpl
```diff
@@ -22,37 +22,34 @@
...
+  "imagePullPolicy": {{ $envALL.Values.images.pull_policy | default "IfNotPresent" | quote }},
```
intlabs: I'm a bit on the fence here, but my gut says that we should not have default values in templates at all if possible.

common/templates/_funcs.tpl
```diff
@@ -22,37 +22,34 @@
...
+  "env": [
+    {
+      "name": "NAMESPACE",
+      "value": "{{ $envALL.Release.Namespace }}"
+    },
+    {
+      "name": "INTERFACE_NAME",
+      "value": "eth0"
+    },
+    {
```
intlabs: Should also include downward API env values and ability to check daemonsets.

heat/templates/deployment-cfn.yaml
```diff
@@ -1,3 +1,5 @@
+{{- $envAll := . }}
+{{- $dependecies := .Values.dependencies.cnf }}
```
intlabs: Typo:
s/cnf/cfn

common/templates/_funcs.tpl
```diff
@@ -22,37 +22,34 @@
...
-      },
-      {
-        "name": "COMMAND",
-        "value": "echo done"
-      }
-    ]
-  }
-]'
+{{- define "dep-check-init-cont" -}}
+{{- $envALL := index . 0 -}}
```
intlabs: I'd keep this as $envAll to maintain consistency

common/templates/_funcs.tpl
```diff
@@ -22,37 +22,34 @@
...
-        "value": "{{  include "joinListWithColon" .jobs }}"
-      },
-      {
-        "name": "COMMAND",
-        "value": "echo done"
-      }
-    ]
-  }
-]'
+{{- define "dep-check-init-cont" -}}
```
intlabs: For consistency could we used '_' instead of '-' as a delineator?

Also rather than overloading the _funcs.tpl I'd be tempted to move this into its own file within the `snippets` directory.

common/templates/_funcs.tpl
DTadrzak: Let's move it to another PR that will cover this and introduce templatable init containers to Neutron.

heat/templates/deployment-cfn.yaml
DTadrzak: Good Catch :)

common/templates/_funcs.tpl
DTadrzak: Yeaha you are right.

common/templates/_funcs.tpl
alanmeadows: Agreed, and note POD_NAME will be required for daemonset checks.  Fine with this in another commit so we can get this in.

v1k0d3n:
All reviews LG.

PR #108, opened 2017-01-12
Removed the ``workers`` configuration from keystone.conf

stannum-l:
The configuration of ``worker`` can be removed for two reasons:

  * In Mitaka, it is two separate parameters:
    ``public_workers`` and ``admin_workers`` under section
    [eventlet_server], as shown in [1].  In master (Ocata),
    these options were removed.
  * In the preferred keystone deployment of using u/wsgi, and not
    eventlet server, this setting does not really take effect - as
    Apache will manage this instead of keystone.

These options can be removed.  Also, removed extra EOL spaces.

[1] https://github.com/openstack/keystone/blob/stable/mitaka/etc/keystone.conf.sample#L678-L696

---
This change is [Reviewable](https://reviewable.io/reviews/att-comdev/openstack-helm/108)

### DIFF
--- a/keystone/templates/etc/_keystone.conf.tpl
+++ b/keystone/templates/etc/_keystone.conf.tpl
@@ -2,7 +2,6 @@
 debug = {{ .Values.misc.debug }}
 use_syslog = False
 use_stderr = True
-workers = {{ .Values.misc.workers }}
 
 [database]
 connection = mysql+pymysql://{{ .Values.database.keystone_user }}:{{ .Values.database.keystone_password }}@{{ include "keystone_db_host" . }}/{{ .Values.database.keystone_database_name }}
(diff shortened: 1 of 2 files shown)


intlabs:
LGTM.

@stannum-l: I see they are totally removed in Newton: https://github.com/openstack/keystone/blob/stable/newton/etc/keystone.conf.sample#L784

wilkers-steve:
:lgtm:

---

Reviewed 2 of 2 files at r1.
Review status: all files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/att-comdev/openstack-helm/108#-:-KaInD8HyenEfME0-V9c:bnfp4nl)*

v1k0d3n:
thanks @stannum-l ! LGTM.

PR #111, opened 2017-01-12
feat: Horizon Add development setting, exposing the service via a NodePort.

intlabs:
This commit adds a `development` option similar to MariaDB. It simply exposes Horizon using a NodePort@31537/tcp to make is a bit more user-friendly to get going in Minikube.

---
This change is [Reviewable](https://reviewable.io/reviews/att-comdev/openstack-helm/111)


v1k0d3n:
@intlabs these are good changes that we'd like to see. if you have a moment to refactor, i'll pull them in.

v1k0d3n:
this got left out on an island, and there is an issue/PR that will supersede this PR. in fact #258 should've referenced this previous work. @intlabs what are your thoughts? i'll leave it up to you, and you can coordinate with @wilkers-steve on what you guys are deciding to do here.

wilkers-steve:
@v1k0d3n the approach for nodeports proposed in PR #261 conflicts with how it's handled for Horizon here. I'm a fan of using the approach in #261 because it doesn't tie nodeports directly to development mode. If we want that to be the case, we should address it in the node port PR and propose a tree structure that would scope them under development.

v1k0d3n:
@wilkers-steve of course...this was just a quick PR to address having horizon exposed for development. i just don't like closing peoples PR's without getting their feedback first. ;)

intlabs:
This has been superseded a number of times, by various methods ;) I'm good with closing this, cheers @v1k0d3n

PR #113, opened 2017-01-12
Update Heat to mount policy at engine pods and set client endpoints

intlabs:
The initial heat commit produced a working API, but did not configure the clients correctly. This commit rectifies that.

---
This change is [Reviewable](https://reviewable.io/reviews/att-comdev/openstack-helm/113)

### DIFF
--- a/heat/values.yaml
+++ b/heat/values.yaml
@@ -108,7 +108,7 @@ resources:
     workers: 8
 
 misc:
-  debug: false
+  debug: true
 
 secrets:
   keystone_admin:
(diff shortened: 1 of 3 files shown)


heat/values.yaml
```diff
@@ -108,7 +108,7 @@ resources:
     workers: 8
 
 misc:
-  debug: false
+  debug: true
```
alanmeadows: Minor nit, normally this defaults to false in other charts.

v1k0d3n:
Reviews look good.

PR #114, opened 2017-01-13
Keystone: Configmap Updates

intlabs:
This PR updates the Keystone Chart:
 * Loads all requires configmaps into container for (apache based) image agnostic deployment
 * Gracefully stops apache when terminating pods, allowing transactions to complete.
 * Removes Kolla specific logic for Keystone Endpoint bootstrapping

---
This change is [Reviewable](https://reviewable.io/reviews/att-comdev/openstack-helm/114)


keystone/templates/etc/_wsgi-keystone.conf.tpl
```diff
@@ -1,6 +1,9 @@
 Listen {{ .Values.network.ip_address }}:{{ .Values.network.port.public }}
 Listen {{ .Values.network.ip_address }}:{{ .Values.network.port.admin }}
 
+LogFormat "%h %l %u %t \"%r\" %>s %b \"%{Referer}i\" \"%{User-Agent}i\"" combined
+LogFormat "%{X-Forwarded-For}i %l %u %t \"%r\" %>s %b \"%{Referer}i\" \"%{User-Agent}i\"" proxy
```
intlabs: By setting up logging like this we can log requests appropriately from both internal and external clients when behind an ingress controller.

keystone/templates/bin/_init.sh.tpl
```diff
@@ -2,5 +2,20 @@
...
+  -m mysql_db -a "login_host='{{ include "keystone_db_host" . }}' \
+                  login_port='{{ .Values.database.port }}' \
+                  login_user='{{ .Values.database.root_user }}' \
+                  login_password='{{ .Values.database.root_password }}' \
+                  name='{{ .Values.database.keystone_database_name }}'"
+
+ansible localhost -vvv \
+  -m mysql_user -a "login_host='{{ include "keystone_db_host" . }}' \
+                    login_port='{{ .Values.database.port }}' \
+                    login_user='{{ .Values.database.root_user }}' \
```
intlabs: In a future PR I'd really like to review this logic, as writing the both the cloud-admin, and MariaDB root password to a configmap does not seem the best idea.

The simplest approach would simply be to use a secret for this entire file instead, though loading the admin credentials via environment variables (again sourced from sperate K8s Secrets) initially seems the most elegant and appropriate methodology.

keystone/templates/deployment.yaml
```diff
@@ -61,26 +61,62 @@ spec:
           ports:
             - containerPort: {{ .Values.network.port.public }}
             - containerPort: {{ .Values.network.port.admin }}
+          lifecycle:
```
alanmeadows: For clarity, I think we should include a terminationGracePeriodSeconds setting along with this and include a values.yaml setting setting this to 30 seconds (what will happen even if we dont set it) for clarity and the potential to override.  Not necessary for this PR though but I would like to see it when we make a pass for graceful termination pattern across all charts.

alanmeadows:
@intlabs This became invalidated during a few merges, can you rebase?

PR #115, opened 2017-01-13
Keystone: Use Yaodu Ubuntu Image

intlabs:
This PR proposes the use of the Keystone Image from the Yaodu Project.

For more information about the image please refer to:
 * https://hub.docker.com/r/yaodu/keystone/
 * https://github.com/yaodu/docker-keystone

---
This change is [Reviewable](https://reviewable.io/reviews/att-comdev/openstack-helm/115)

PR #116, opened 2017-01-13
Heat: Use Yaodu Ubuntu Image

intlabs:
This PR proposes the use of the Heat Image from the Yaodu Project.

For more information about the image please refer to:
 * https://hub.docker.com/r/yaodu/heat/
 * https://github.com/yaodu/docker-heat

---
This change is [Reviewable](https://reviewable.io/reviews/att-comdev/openstack-helm/116)


heat/values.yaml
```diff
@@ -17,14 +17,14 @@ labels:
...
   db_init: quay.io/stackanetes/stackanetes-kolla-toolbox:newton
-  db_sync: docker.io/kolla/ubuntu-source-heat-api:3.0.1
-  ks_user: quay.io/stackanetes/stackanetes-kolla-toolbox:newton
-  ks_service: quay.io/stackanetes/stackanetes-kolla-toolbox:newton
-  ks_endpoints: quay.io/stackanetes/stackanetes-kolla-toolbox:newton
-  api: docker.io/kolla/ubuntu-source-heat-api:3.0.1
-  cfn: docker.io/kolla/ubuntu-source-heat-api:3.0.1
-  cloudwatch: docker.io/kolla/ubuntu-source-heat-api:3.0.1
-  engine: docker.io/kolla/ubuntu-source-heat-engine:3.0.1
+  db_sync: docker.io/yaodu/heat:ubuntu
```
alanmeadows: Same comment here as in #115 - how does this allow openstack release targeting? I am assuming there is only one release of yaodu right now, but the naming seems short sighted?

heat/values.yaml
```diff
@@ -108,7 +108,7 @@ resources:
     workers: 8
 
 misc:
-  debug: false
+  debug: true
```
alanmeadows: We normally default to false.

heat/values.yaml
SamYaple: the naming follows [docker upstream](https://github.com/docker-library/official-images#tags-and-aliases) for better or worse.

Additionally, I have this https://etherpad.openstack.org/p/yaodu-tagging that is open to suggestions

v1k0d3n:
i'd like to potentially use the slimmer yaodu images (default) as a complimentary project once in openstack proper, but i believe that we're going to let this PR drop (close) prior to the repo import.

cc: @intlabs @SamYaple

intlabs:
@v1k0d3n that works for me - don't worry, I'll be back pushing these again when both projects are at a level of maturity that makes sense to do so :)

Issue #100, opened 2017-01-10
feat: Extend rabbitmq chart with autocluster to support multiple rabbitmq instances


alanmeadows:
See https://github.com/aweber/rabbitmq-autocluster

We should investigate leveraging the DNS backend, since this should be supported natively with a kubernetes service, as opposed to using etcd.

For a working example, see https://github.com/openstack/fuel-ccp-rabbitmq - however this example uses etcd and would prefer to validate the DNS approach works.

DTadrzak:
One guy from Intel is working to improve rabbitmq chart.

alanmeadows:
Great news @DTadrzak.  Once we tackle rabbit and memcached, the resiliency will be significantly improved.

intlabs:
@SamYaple may also have some good insight on this.

SamYaple:
I use rabbitmq-clusterer. It works well. autoclusterer is great if you have consul/etcd or something that it supports already setup. clusterer doesn't have any external deps, but requirements better config management from a deployment tool.

Just for more info, rabbitmq 3.7.0 has most of the features from both of these tools baked in.

DTadrzak:
AFAIK @ss7pro's solution is based on etcd.

SamYaple:
the autoclusterer is a good choice, yes. In you situation, I would go with autoclusterer.

janwillies:
We are running RabbitMQ with Kubernetes-based discovery and it's not great tbh. While I agree it's the cleanest approach there's lots of custom work to be done. You need to run an unreleased version of autocluster (HEAD basically). You need to [hack-in](https://github.com/aweber/rabbitmq-autocluster/issues/92#issuecomment-245381003) a custom restart because autocluster doesn't like being started together with other plugins (or vice versa). Then the RabbitMQ cluster is happily partitioning on the first occasion and you need to handle that somehow. While it sounds easy on the Mirantis [blog post](https://www.mirantis.com/blog/clustered-rabbitmq-kubernetes/), `autocluster:cluster_health_check_report()` is only available in their [fork](https://github.com/Mirantis/rabbitmq-autocluster/commit/5fee57752a0788bd2358d3f09eae76d4da67f039) and tailored towards the etcd backend.

alanmeadows:
@janwillies This is useful real-world information.  

I am hearing that the the dream of a "native" DNS-backed autoclustering scenario is the part of this issue that should be put on ice for the moment.

[...]

PR #123, opened 2017-01-17
Add rabbitmq deployment with HA.

ss7pro:
This works is based on http://github.com/openstack-fuel-ccp-rabbitmq and it mostly adjust this solution
to a helm deployment.

This is works in progress, but we need to figure out if we really need:

1) HA etcd deployment. In this case etcd is used just for a coordination purposes. Fuel-ccp is having non-HA etcd and relaying just on deployment controller. Stackanetes is using etcd in HA but this is mostly due to kpm sharding feature so it was easy to build etcd cluster without statefullset. 

2) Having stuff in StateFullSet means that we can upgrade that component (K8S does not support upgrades on StateFullSet).

---
This change is [Reviewable](https://reviewable.io/reviews/att-comdev/openstack-helm/123)


DTadrzak:
Related to #100

rabbitmq/templates/deployment.yaml
```diff
@@ -27,12 +25,15 @@ spec:
...
+        - name: rabbitmq-etc
+          configMap:
+            name: rabbitmq-etc
       containers:
         - name: rabbitmq
           image: {{ .Values.images.rabbitmq }}
           command:
-            - bash
-            - /tmp/start_rabbitmq.sh
+            - sleep
```
DTadrzak: leftover

rabbitmq/values.yaml
```diff
@@ -27,4 +27,19 @@ network:
     management: '15672'
 
 images:
-  rabbitmq: "rabbitmq:3-management"
+  rabbitmq: "quay.io/stackanetes/stackanetes-glance-api:newton"
```
DTadrzak: ?

rabbitmq/values.yaml
```diff
@@ -27,4 +27,19 @@ network:
...
+
+enabled_plugins:
+  - autocluster
```
alanmeadows: Should we support some visualization plugin by default as well - may require an additional port exposed.

etcd/templates/deployment.yaml
```diff
@@ -0,0 +1,36 @@
+apiVersion: extensions/v1beta1
+kind: Deployment
+metadata:
+  name: etcd
+spec:
```
alanmeadows: It seems that certainly etcd, and rabbit should take advantage of scheduling to ensure proper node placement for resiliency - this supports both large environments and collapsed (all-in-one) environments:

https://github.com/att-comdev/openstack-helm/blob/master/mariadb/templates/deployment.yaml#L18-L38

alanmeadows:
@DTadrzak @ss7pro do you know where we are with this? I'd like to get this in before the PTG.

rabbitmq/values.yaml
ss7pro: Which plugin you mean ? We don't have any one tested.

etcd/templates/deployment.yaml
ss7pro: Hey etcd is single instance deployment without HA (as it's not needed). I've added anit-affinity to rabbitmq deployment.

etcd/templates/deployment.yaml
```diff
@@ -0,0 +1,36 @@
...
+      containers:
+        - name: etcd
+          image: {{ .Values.images.etcd }}
+          imagePullPolicy: {{ .Values.images.pull_policy }}
+          command:
+            - etcd
+            - --listen-client-urls
+            - http://0.0.0.0:{{ .Values.network.port }}
+            - --advertise-client-urls
+            - http://etcd:{{ .Values.network.port }}
```
DTadrzak: etcd service-name value should be configurable

rabbitmq/templates/deployment.yaml
ss7pro: done

rabbitmq/values.yaml
ss7pro: done

ss7pro:
@alanmeadows DNS and native K8S autoclustering is full of multiple race conditions and possible split brain situations. Mirantis invested significant resources into etcd based clustering and this implementation is bulletproof.

ss7pro:
@alanmeadows Have a look at this PR now, it's ready. I'll open few more issues as we should move anti-affinity to common package and init-container inclusion should be modified as in K8S 1.6 it's no longer defined using annotations (init containers are moved out of beta).

etcd/templates/deployment.yaml
ss7pro: done.

rabbitmq/templates/etc/_enabled_plugins.tpl
```diff
@@ -0,0 +1,4 @@
+[{{- range .Values.enabled_plugins -}}
```
DTadrzak: instead of that please use joinlistwithcomma(https://github.com/att-comdev/openstack-helm/blob/master/common/templates/_funcs.tpl#L1) 
like here:
https://github.com/att-comdev/openstack-helm/blob/d18f333bbf6da5862c98a446e7e8f84a67a206c8/neutron/templates/etc/_ml2-conf.ini.tpl#L3

etcd/requirements.yaml
```diff
@@ -0,0 +1,4 @@
+dependencies:
+  - name: common
+    repository: http://localhost:8879/charts
+    version: 0.1.0
```
DTadrzak: add etcd

ss7pro: done

etcd/templates/deployment.yaml
ss7pro: done

rabbitmq/templates/etc/_enabled_plugins.tpl
ss7pro: done

DTadrzak:
lgtm

etcd/requirements.yaml
wilkers-steve: Yep.  This is the current issue with the Travis jobs failing currently

docker/rabbitmq/Dockerfile.j2
```diff
@@ -0,0 +1,40 @@
...
+# `cp` is needed until https://github.com/rabbitmq/rabbitmq-server/pull/1016 is merged
+RUN curl -Lo /tmp/rabbitmq-autocluster.deb {{ url_autocluster_provider }}/rabbitmq-autocluster_{{ autocluster_version }}_all.deb \
+  && dpkg -i /tmp/rabbitmq-autocluster.deb \
+  && cp -v /usr/lib/rabbitmq/plugins/*.ez /usr/lib/rabbitmq/lib/rabbitmq_server-*/plugins/
+
+
+RUN usermod -a -G rabbitmq rabbitmq \
+    && chown -R rabbitmq: /var/lib/rabbitmq /var/log/rabbitmq /etc/rabbitmq
+
+USER rabbitmq
```
alanmeadows: Thinking this must be a left-over from development -- no other charts include Dockerfiles.

openstack/requirements.yaml
```diff
@@ -7,7 +7,7 @@ dependencies:
     version: 0.1.0
   - name: rabbitmq
     repository: http://localhost:8879/charts
-    version: 0.1.0
+    version: 0.2.0
```
alanmeadows: Not a nit just informational, I think we should delete the openstack parent chart.  It is not well maintained and not really a path I think we are advertising.

rabbitmq/requirements.yaml
```diff
@@ -2,3 +2,6 @@ dependencies:
   - name: common
     repository: http://localhost:8879/charts
     version: 0.1.0
+  - name: etcd
```
alanmeadows: This would be the very first chart that has any sort of dependency based on our pattern -- helm dependency management isn't where we think it should be to start using this.  Instead, we are eyeing independent charts with [https://github.com/att-comdev/armada](something on top to orchestrate installation).

I believe that etcd should be its own chart, maintained to the best of its ability to be a resilient etcd service.  This chart (rabbit) is simply a user of that service.

Remember that as soon as this becomes a helm dependency helm will install it.  That isn't a problem for macro-only charts like common, but for charts with actual resources like etcd, if this is a dependency of any other chart in the same namespace, it will try to install etcd twice which I don't think is what you want. 

Why wouldn't this be like any other pattern we have, for instance, how glance does need keystone, but it is not a dependency of glance. 

Some users may want this rabbitmq but have their own etcd installation or want it on baremetal or ...

rabbitmq/templates/deployment.yaml
```diff
@@ -18,6 +20,27 @@ spec:
         app: rabbitmq
       annotations:
         configmap-bin-hash: {{ tuple "configmap-bin.yaml" . | include "hash" }}
+        configmap-etc-hash: {{ tuple "configmap-etc.yaml" . | include "hash" }}
```
alanmeadows: Nice catch!

rabbitmq/values.yaml
```diff
@@ -27,7 +27,7 @@ upgrades:
     max_surge: 3
 auth:
   default_user: openstack
-  default_pass: password
+  default_password: password
```
alanmeadows: Minor nit, either _password or _pass everywhere

rabbitmq/values.yaml
```diff
@@ -37,4 +37,26 @@ network:
     management: '15672'
 
 images:
-  rabbitmq: "rabbitmq:3-management"
+  rabbitmq: "quay.io/stackanetes/stackanetes-rabbitmq:0.6.1.950-1"
```
alanmeadows: We've got a bit of a hodge-podge of images being used, some from Stackanetes, some from Kolla, at some point we need to standardize.  Is there no alternative "official" upstream version of this same image? I'm guessing this is your hosted version of the image built from the fuel-ccp Docker image?

etcd/templates/deployment.yaml
ss7pro: @v1k0d3n https://github.com/coreos/etcd-operator/blob/master/README.md#limitations etcd-operator does not work well on K8S 1.5.X. I would opt to go with single instance as of now and switch to etcd-operator when it's fixed.

docker/rabbitmq/Dockerfile.j2
ss7pro: ok, I've removed it.

docker/rabbitmq/sources.list.debian
```diff
@@ -0,0 +1,4 @@
+# Rabbitmq repo
+deb http://www.rabbitmq.com/debian/ testing main
+# Jessie backports
+deb http://ftp.debian.org/debian jessie-backports main
```
ss7pro: done

etcd/requirements.yaml
ss7pro: done

rabbitmq/requirements.yaml
ss7pro: done

rabbitmq/values.yaml
ss7pro: done

rabbitmq/values.yaml
ss7pro: Kolla is not building rabbitmq with custom Miranti's autocluster plugin. Mirantis invested significant resourced into making autocluster plugin with etcd backend bulletproof (leader election race condition, split brain avoidance, cluster recovery). Their flavor undergo significant destructive testing to confirm it's production readiness.

v1k0d3n: @ss7pro going a bit further on @alanmeadows point; this project will need standardize on images (as Alan said), but we want to be flexible about the framework we're providing as well. keeping in mind that we're trying to abstract from the runtime (using Kubernetes natives), i'd like to see any Dockerfile/acbuild references or container build templates in complimentary projects and outside of the openstack-helm project specifically.

alanmeadows: @ss7pro I think I blurred my request: does Mirantis make their image publicly available so we can point to the original source rather then our own image based on their unchanged Dockerfile?

mredlars:
Tested on 2 of our production environments. Works like a charm, thumbs up!

rabbitmq/values.yaml
ss7pro: @alanmeadows I've changed url to mirantis registry.

ss7pro:
I just rebased this PR and adjusted changes to be aligned with recently introduced helm-toolkit.

ss7pro:
Merged recent fuel-ccp-rabbitmq bugfix: https://github.com/openstack/fuel-ccp-rabbitmq/commit/c524d3d310e2f3f0c7fbe1b8030a4c0b72eac398.

rabbitmq/values.yaml
```diff
@@ -37,4 +37,26 @@ network:
...
+  rabbitmq: "registry.mcp.fuel-infra.org/mcp/rabbitmq:ocata-unstable"
+  dep_check: "quay.io/stackanetes/kubernetes-entrypoint:v0.1.1"
+  pull_policy: "IfNotPresent"
+
+enabled_plugins:
+  - autocluster
+
+erlang_cookie: openstack-cookie
+
+etcd:
```
alanmeadows: I would like to take this under endpoints: at some point when both memcache and databases are handled there with lookup routines for all three.

rabbitmq/values.yaml
```diff
@@ -37,4 +37,26 @@ network:
     management: '15672'
 
 images:
-  rabbitmq: "rabbitmq:3-management"
+  rabbitmq: "registry.mcp.fuel-infra.org/mcp/rabbitmq:ocata-unstable"
```
alanmeadows: Without looking, I'm guessing ocata is the only choice

rabbitmq/values.yaml
```diff
@@ -3,7 +3,7 @@
 # Declare name/value pairs to be passed into your templates.
 # name: value
 
-replicas: "1" # this must be quoted to deal with atoi
+replicas: "3" # this must be quoted to deal with atoi
```
alanmeadows: We should remove the quotes, no atoi in use anymore

etcd/Chart.yaml
```diff
@@ -0,0 +1,3 @@
+description: Chart for etcd
+name: etcd
+version: 0.2.0
```
alanmeadows: Regarding versions, I think we should roll them all up as part of tagging/branching and do it as a snapshot, otherwise we'll end up with a scattered mix - so 0.1.0 here.

openstack/requirements.yaml
alanmeadows: Again, think 0.1.0 is more appropriate still until the next tag

rabbitmq/Chart.yaml
```diff
@@ -1,4 +1,4 @@
 apiVersion: v1
 description: A Helm chart for Kubernetes
 name: rabbitmq
-version: 0.1.0
+version: 0.2.0
```
alanmeadows: Same

rabbitmq/values.yaml
ss7pro: yep, this is the most stable one.

rabbitmq/values.yaml
ss7pro: done

etcd/Chart.yaml
ss7pro: done

rabbitmq/Chart.yaml
ss7pro: done

alanmeadows: @ss7pro I think this went the opposite way ;-)

I was suggesting 0.1.0 everywhere - and when we move _all_ charts to the next version (presumably 0.2.0) close to Feb 28 we'd do it as one clean sweep.

ss7pro: This is comment to outdated diff, right now it's 0.1.0 everywhere.

etcd/requirements.yaml
```diff
@@ -0,0 +1,4 @@
+dependencies:
+  - name: common
```
alanmeadows: Not sure how we missed this or how the tests are passing; this should be helm-toolkit, not common

ss7pro: ok, fixed

Makefile
```diff
@@ -1,9 +1,9 @@
...
+.PHONY: ceph bootstrap mariadb etcd postgresql keystone memcached rabbitmq helm-toolkit openstack neutron nova cinder heat maas all clean
 
 B64_DIRS := helm-toolkit/secrets
 B64_EXCLUDE := $(wildcard helm-toolkit/secrets/*.b64)
 
-CHARTS := ceph mariadb postgresql rabbitmq memcached keystone glance horizon neutron nova cinder heat maas openstack
+CHARTS := ceph mariadb etcd postgresql rabbitmq memcached keystone glance horizon neutron nova cinder heat maas openstack
 TOOLKIT_TPL := helm-toolkit/templates/_globals.tpl
 
 all: helm-toolkit ceph bootstrap mariadb postgresql rabbitmq memcached keystone glance horizon neutron nova cinder heat maas openstack
```
alanmeadows: This is why the tests are not erroring on etcd build - etcd is missing from this line

ss7pro: added.

PR #130, opened 2017-01-21
Add daemonset support to Kubernetes Entrypoint init-container

intlabs:
This commit adds support for daemonsets to the Kubernetes-Entrypoint init-container.

It also makes two other changes:
 * Moves the entrypoint container manifest snippet to its own file to reducle loading on the _funcs.tpl file
 * changes `dep-check-init-cont` to `dep_check_init_cont` to match the formatting of other defines used in OpenStack Helm

---
This change is [Reviewable](https://reviewable.io/reviews/att-comdev/openstack-helm/130)

PR #131, opened 2017-01-21
Neutron: Move to use init-containers for dependency checking

intlabs:
This commit brings Neutron inline with other services, by moving the dependency checks to an init-container.

It builds on the commit in https://github.com/att-comdev/openstack-helm/pull/130.

As a result that PR should be reviewed first, once (and if) it is accepted, only https://github.com/att-comdev/openstack-helm/pull/131/commits/2b09f1deb9bc307afe01803b9286149a1f759a0f is relevant in this PR.

---
This change is [Reviewable](https://reviewable.io/reviews/att-comdev/openstack-helm/131)


intlabs:
Superceeded by: https://github.com/att-comdev/openstack-helm/pull/138

PR #134, opened 2017-01-21
Update endpoint function to accomodate empty paths

intlabs:
This commit allows the endpoint function to cope cleanly with services that do not specify a path.

---
This change is [Reviewable](https://reviewable.io/reviews/att-comdev/openstack-helm/134)


intlabs:
Will be superseded shortly

PR #135, opened 2017-01-21
Replace Ansible with OpenStack Client for Glance Keystone Jobs

intlabs:
Replace Ansible with OpenStack Client for Glance

This commit removes the dependency on Ansible for the creation and management of Keystone Users and Endpoints.

It makes use of the OpenStack Client which is part of the standard requirements for all OpenStack Services.
The OpenStack Client is therefore included every OpenStack Service Container regardless of origin.

It builds on the commits in the following PR, which should be reviewed prior to this:
 * https://github.com/att-comdev/openstack-helm/pull/130
 * https://github.com/att-comdev/openstack-helm/pull/134

The new commit introduced in this review is: https://github.com/att-comdev/openstack-helm/pull/135/commits/12e13cea87af63933f7223bbc505c3e9ebe134b5

---
This change is [Reviewable](https://reviewable.io/reviews/att-comdev/openstack-helm/135)


intlabs:
Will be superseded shortly

PR #140, opened 2017-01-22
Replace Ansible with SQLAlchemy/PyMySQL for database creation and management.

intlabs:
Removes the dependency on Ansible for the creation and management of databases for OpenStack Services. It adds a small python script that utilizes PyMySQL via SQLAlchemy, two libraries that are present in every OpenStack Service Container that consumes a database regardless of origin.

---
This change is [Reviewable](https://reviewable.io/reviews/att-comdev/openstack-helm/140)


intlabs:
this has been superseded by: https://github.com/att-comdev/openstack-helm/pull/206

PR #139, opened 2017-01-22
Standardize DB Job names

intlabs:
Standardize the DB creation job name to <service>-db-init for all services.

Builds on the following which should be reviewed/merged first:
 * https://github.com/att-comdev/openstack-helm/pull/137
 * https://github.com/att-comdev/openstack-helm/pull/138

---
This change is [Reviewable](https://reviewable.io/reviews/att-comdev/openstack-helm/139)

PR #142, opened 2017-01-22
Refactor Keystone Endpoint Functions and Bootstrap Jobs For services

intlabs:
Removes the dependency on Ansible for the creation and management of Users, Services and Endpoints for OpenStack Services.

As part of this PR the Endpoint functions are updated from being based around service name to service type. This brings OpenStack Helm inline with other OpenStack Deployment tools and makes the management of versioned endpoints possible.

Builds on the following which should be reviewed/merged first:
 * https://github.com/att-comdev/openstack-helm/pull/137
 * https://github.com/att-comdev/openstack-helm/pull/138
 * https://github.com/att-comdev/openstack-helm/pull/139
 * https://github.com/att-comdev/openstack-helm/pull/140

The commits specific to this PR start at: "Update Common Lib Endpoint Functions" (currently https://github.com/att-comdev/openstack-helm/pull/142/commits/5000b66120f941c5a6bdd234bc546b5a964b15b6) If reviewed individually they are a lot less scary than would appear from the overall size of this PR.

---
This change is [Reviewable](https://reviewable.io/reviews/att-comdev/openstack-helm/142)


common/templates/_endpoints.tpl
```diff
@@ -2,136 +2,44 @@
...
+{{- define "endpoint_name_lookup" -}}
+{{- $type := index . 0 -}}
 {{- $context := index . 1 -}}
-{{- $nameNorm := $name | replace "-" "_" }}
-{{- $endpointMap := index $context.Values.endpoints $nameNorm }}
-{{- $endpointType := index $endpointMap "type" }}
-{{- $endpointType | quote -}}
+{{- $endpointMap := index $context.Values.endpoints $type }}
+{{- $endpointName := index $endpointMap "name" }}
+{{- $endpointName | quote -}}
```
intlabs: the quote here needs to be removed

common/templates/_endpoints.tpl
```diff
@@ -2,136 +2,44 @@
...
+{{- $endpointMap := index $context.Values.endpoints $type }}
 {{- $fqdn := $context.Release.Namespace -}}
 {{- if $context.Values.endpoints.fqdn -}}
 {{- $fqdn := $context.Values.endpoints.fqdn -}}
 {{- end -}}
 {{- with $endpointMap -}}
 {{- $endpointScheme := .scheme }}
-{{- $endpointHost := index .hosts $endpoint | default .hosts.default}}
+{{- $endpointHost := index .hosts $endpoint | default .hosts.default }}
 {{- $endpointPort := index .port $port }}
```
alanmeadows: We should be able to default here, shouldn't we? The port is in `endpoints:`

common/templates/_endpoints.tpl
```diff
@@ -2,136 +2,44 @@
...
-{{- $endpointPath := .path }}
-{{- printf "%s://%s.%s:%1.f%s" $endpointScheme $endpointHost $fqdn $endpointPort $endpointPath  | quote -}}
+{{- $endpointPath := .path | default "" }}
+{{- printf "%s://%s.%s:%1.f%s" $endpointScheme $endpointHost $fqdn $endpointPort $endpointPath -}}
 {{- end -}}
 {{- end -}}
 
 
 #-------------------------------
```
alanmeadows: Couldn't the map live in common so it doesn't have to be unnecessarily duplicated in every charts `endpoints:` - it won't change much once its defined.

glance/templates/secret-keystone-user.env.yaml
```diff
@@ -0,0 +1,20 @@
+apiVersion: v1
```
alanmeadows: This whole thing seems generic - couldn't we pass .Values.keystone to a common template for this type of secret?

glance/templates/secret-keystone-admin.env.yaml
```diff
@@ -0,0 +1,20 @@
+apiVersion: v1
```
alanmeadows: This whole thing seems generic - couldn't we pass .Values.keystone to a common template for this type of secret?

glance/templates/etc/_glance-api.conf.tpl
```diff
@@ -8,27 +8,29 @@ workers = {{ .Values.misc.workers }}
...
 [database]
 connection = mysql+pymysql://{{ .Values.database.glance_user }}:{{ .Values.database.glance_password }}@{{ .Values.database.address }}/{{ .Values.database.glance_database_name }}
 max_retries = -1
-    
+
 [keystone_authtoken]
-auth_uri = {{ .Values.keystone.auth_uri }}
-auth_url = {{ .Values.keystone.auth_url }}
+auth_version = v3
+auth_url = {{ tuple "identity" "internal" "public" . | include "endpoint_type_lookup_addr" }}
```
alanmeadows: I wonder if there is a better way for us to articulate 5000 vs 35.... (public vs admin) - "identity" "internal" "public" makes me read this line two times before I remember whats going on.

If we adjusted the values, we might be able to make it read - just throwing this out as we are setting a standard here:

`identity internal public_port`

horizon/templates/etc/_local_settings.tpl
```diff
@@ -130,7 +130,7 @@ SESSION_ENGINE = 'django.contrib.sessions.backends.cache'
 CACHES = {
     'default': {
         'BACKEND': 'django.core.cache.backends.memcached.MemcachedCache',
-        'LOCATION': '{{ include "memcached_host" . }}'
+        'LOCATION': '{{ .Values.memcached.host }}:{{ .Values.memcached.port }}'
```
alanmeadows: I think memcache should be an endpoint like everything else.

heat/values.yaml
```diff
@@ -80,6 +79,8 @@ service:
     proto: "http"
 
 database:
```
alanmeadows: I think the database should be an endpoint like everything else.

horizon/values.yaml
```diff
@@ -26,6 +26,10 @@ network:
   node_port: 30000
   enable_node_port: false
 
+memcached:
```
alanmeadows: Same.  See above.

keystone/templates/etc/_keystone.conf.tpl
```diff
@@ -4,7 +4,8 @@ use_syslog = False
 use_stderr = True
 
 [database]
-connection = mysql+pymysql://{{ .Values.database.keystone_user }}:{{ .Values.database.keystone_password }}@{{ include "keystone_db_host" . }}/{{ .Values.database.keystone_database_name }}
+connection = mysql+pymysql://{{ .Values.database.keystone_user }}:{{ .Values.database.keystone_password }}@{{ .Values.database.address }}:{{ .Values.database.port }}/{{ .Values.database.keystone_database_name }}
```
alanmeadows: Same.  And if it was, you can use your new routine to generate this.  Endpoint for dbs just gets expanded to have database username and passwords (path is already akin to dbname)

keystone/templates/secret-db-root.env.yaml
```diff
@@ -0,0 +1,9 @@
+{{- $dbRootConnection := printf "mysql+pymysql://%s:%s@%s:%1.f" .Values.database.root_user .Values.database.root_password .Values.database.address .Values.database.port }}
```
alanmeadows: Same.

alanmeadows:
@intlabs Pete, I'd love to see this taken to completion as its huge for consistency.  If you need to chunk any of it up, I'm happy to help.  If you want, ignore the request to rope in databases and memcache and I'll take that on.

gardlt:
can we merge this pr dep for #188 #197

v1k0d3n:
@gardlt once some changes have been made and the conflicts are resolved...the plan is to pull this in (and even possibly break this up to get it in a big faster).

intlabs:
This is undergoing refactoring, and should not be merged until complete.

intlabs:
This PR is being closed in favor of a series of smaller PR's beginning with https://github.com/att-comdev/openstack-helm/pull/208, with smaller scopes that will be much more manageable to review.

Issue #32, opened 2016-12-12
security: look into restricting access to the kubernetes API (only where needed)


v1k0d3n:
i would like to look at policies that could restrict access from all services having direct access to the API (wherever possible/within reason). perhaps a PoC with something like https://github.com/aporeto-inc/trireme could be used, or possibly calico/romana?

intlabs:
This has been a long ongoing issue with Kubernetes, though I think that enabling ABAC in the cluster would be a big step in the right direction:
 * https://github.com/kubernetes/kubernetes/issues/16779
 * https://github.com/kubernetes/kubernetes/pull/17445

We should also review the security of the Kubernetes Entrypoint container - and possibly look toward updating this with a server - client model, which would be fairly easy to do:
 * 'Server' lives in kube-system namespace, consisting of K8S API watcher, ETCD and nginx containers:
   1. K8s Watcher stores selected job/service status in ETCD.
   2. Nginx provides read-only proxy of data for ETCD to pods, potentially with https basic auth. 
 * 'Client' container could be very simple script that simply curls the Nginx container.

v1k0d3n:
also, i would like our internal security group look at Kubernetes `NetworkPolicy` and calico/canal Advanced `NetworkPolicy`.

A couple areas of interest here:
All charts/manifests should include some form of `NetworkPolicy` (not all addressed today): http://kubernetes.io/docs/user-guide/networkpolicies/

v1k0d3n:
we can track this as a larger issue, but it is unrelated to a release...so the issue is going to be closed for now.

PR #143, opened 2017-01-23
wip: Do Not Merge. Possible alternative dependency checking method.

intlabs:
WIP Prototype of an alternate implementation of dependency checking.

 * Relies on the OpenStack Client / SQLAlchemy (which is included with every OpenStack project as a dependency), so no extra image or binaries are required.
 * Does not interact with the API, so pods can be isolated from it, reducing the potential attack surface upon K8s.
 * Failing rather than waiting, so operators gain visibility of pods/services that require attention.
 * Checking service operation directly, rather than relying on status checks, so we get confirmation that not only are the required services reporting as available but that they are also actually serving requests, and accessible by the pod that needs to consume it.
 * Removing the coupling from K8s services and job history allows OpenStack Helm to be simply deployed in many more configurations than previously possible, including federated clusters, brownfield sites and/or in combination with external services, whilst retaining dependency checking.

Currently, this is very much a WIP, and only supports DB and Keystone-related functions, but the pattern can easily be expanded to cover sockets and other services as required.

The commit that is relevant in this PR is https://github.com/att-comdev/openstack-helm/commit/c49edfbd2c4613f0325805721c3c781b77ac1976 (particularly from [here](https://github.com/att-comdev/openstack-helm/commit/c49edfbd2c4613f0325805721c3c781b77ac1976#diff-ef3489baee269a4668d44fa7e94d0b38) onwards), which uses Cinder as an example.
Sample output from this prototype is here: 
* https://gist.github.com/intlabs/06541d4aed6d346d8ed7200791f87375#file-cinder-openstack-output
* https://gist.github.com/intlabs/06541d4aed6d346d8ed7200791f87375#file-cinder-openstack-helm-entrypoint-L912-L1028

This Prototype/Proposal is potentialy relevant to the discussions in https://github.com/att-comdev/openstack-helm/issues/88 and https://github.com/att-comdev/openstack-helm/issues/32.

---
This change is [Reviewable](https://reviewable.io/reviews/att-comdev/openstack-helm/143)


alanmeadows:
One thought that occurred to me - this would also need to account for validating the _schema_ is in place as well.  Even better if it validates the _correct_ schema is in place but likely that would have to be future.

In other words, this would OK if the databas existed but dbsync was still in flight.  This can take 5+ minutes on a nova/neutron install as it runs through the history of patches.  The fear would be we would allow things to continue, to connect, only to find them crash as they tried to access tables that don't exist.

intlabs:
@alanmeadows: Though we have discussed this offline I'll add some comments here for posterity:

I made a start at the db schema checking using cinder as a guinea pig:
 * [values.yaml](https://github.com/portdirect/openstack-helm/blob/7296df19e0e5aab5bed7f6ad94449aa930b138d8/cinder/values.yaml#L110-L113)
 * [init-container-snippet](https://github.com/portdirect/openstack-helm/blob/7296df19e0e5aab5bed7f6ad94449aa930b138d8/common/templates/snippets/_k8s_init_dep_check_db.tpl#L60)
 * [usege in a pod template](https://github.com/portdirect/openstack-helm/blob/7296df19e0e5aab5bed7f6ad94449aa930b138d8/cinder/templates/deployment-api.yaml#L27)

To complete this, we will need to get the schema version for each OpenStack release and then compare the value returned from the db version command.

Over the next few days I'll get this a bit more PR worthy, most likely by starting a new PR that breaks the concepts introduced here into more manageable chunks.

alanmeadows:
I continue to support this change.  One comment:

The stab above would suggest as long as db version exits successfully, we would be good to go.  This is an improvement over a simple auth check to the database.  But it does differ from our discussion of comparing it to a specific target version.  Perhaps that is a next evolution I am simply overlooking?

It's a good stop gap but the one challenge with this as is (versus what we discussed, checking for a specific version pinned to the target release) would be during upgrades.  This would likely return success while a second (post-install, during upgrade) db_sync is in progress on an existing schema taking it to the target version.  The old "legacy" approach would of protected against this because an upgrade would of required the end user delete the db sync job before upgrading and changing the container image for that job (its own problem, I'll admit), thus ensuring entrypoint checks on the job completing would hang until it finishes.  If we want to replace that I think we need to make sure we don't move backward in capabilities.

v1k0d3n:
this work got left on an island too, and i don't want to lose any of the effort (especially the great conversation). perhaps we can approach this work in phases or create issues for the parts of this work we don't want to lose if this PR were to drop. thoughts @intlabs?

intlabs:
@v1k0d3n This PR is very hard to break up in its current form. But I'm very keen on continuing this work as it will be vital for cross-cluster operation and interaction with externally managed infrastructure services. Once the DB work in https://github.com/att-comdev/openstack-helm/pull/206 is merged I'll be able to make meaningful progress on this topic again.

v1k0d3n:
ok sounds good. @intlabs we'll work on this tomorrow ;)

wilkers-steve:
Thanks for this work @intlabs.  We'll file a blueprint once we're in OpenStack to track this work going forward.  Until then, we'll leave this PR open for you to reference after that point.

v1k0d3n:

PR #148, opened 2017-01-25
Replace maas systemd service with kubernetes job

larryrensing:
also added configurable maas admin credentials

---
This change is [Reviewable](https://reviewable.io/reviews/att-comdev/openstack-helm/148)


maas/templates/bin/_import-boot-resources.sh.tpl
```diff
@@ -0,0 +1,33 @@
+#!/bin/sh
+
+set -ex
+
+function check_for_download {
```
alanmeadows: This is a nice clean check.  I like it.

maas/templates/bin/_import-boot-resources.sh.tpl
```diff
@@ -0,0 +1,33 @@
...
+        sleep 60
+        exit 1
+    else
+        echo 'Boot resources found'
+        exit 0
+    fi
+
+}
+
+maas-region local_config_set \
```
alanmeadows: This is actually an improvement over overlaying a configmap because this file has so few settings.

maas/templates/bin/_import-boot-resources.sh.tpl
```diff
@@ -0,0 +1,33 @@
...
+        --database-host "{{ .Values.db_service_name }}.{{ .Release.Namespace}}" \
+        --database-name "{{ .Values.database.db_name }}" \
+        --database-user "{{ .Values.database.db_user }}" \
+        --database-pass "{{ .Values.database.db_password }}" \
+        --maas-url "http://{{ .Values.ui_service_name }}.{{ .Release.Namespace }}:80/MAAS"
```
alanmeadows: The port should probably be configurable if you also add support for it on the apache end.

maas/templates/bin/_import-boot-resources.sh.tpl
```diff
@@ -0,0 +1,33 @@
+#!/bin/sh
+
+set -ex
+
+function check_for_download {
+
+    if maas {{ .Values.credentials.admin_username }} boot-resources read | grep -q '\[\]';
+    then
+        echo 'Did not find boot resources.  Will try again'
+        sleep 60
```
alanmeadows: What we have noticed, is that this is not as idempotent as one would like.  What could potentially happen here is that we attempt to import, that succeeds (as in it starts and is doing stuff), but it does not complete enough to provide an image in the list by the time you test.  The sleep 60 is a start, but you can get into an infinite cycle of kicking it off, testing (and failing), and then issuing another kick.  

Also, because you kick off an image fetch, then sleep for 10 seconds, then check, really your only giving the fetch 10 seconds to complete.  If it fails, _then_ we wait 60 seconds before starting it all over again.

I would instead suggest:

- kick off image fetch
- wrapped_in_a_forgiving_timeout_such_as_15_minutes_from_values_so_operator_can_tune:
-- every 10 seconds, check for images and if we find them, exit 0.  If not, print that we are still waiting.
-- if 15 minutes wrapped timeout expires, exit 1 to restart the whole process.

maas/templates/deploy-region.yaml
```diff
@@ -13,7 +13,7 @@ spec:
           {
             "name": "init",
             "image": "{{ .Values.images.maas_region }}",
-            "imagePullPolicy": "Always",
+            "imagePullPolicy": "{{ .Values.images.pull_policy }}",
```
alanmeadows: Nice catch, Larry.

maas/templates/job-import.yaml
```diff
@@ -0,0 +1,72 @@
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: region-import-resources
+spec:
+  template:
+    metadata:
+      annotations:
+        pod.beta.kubernetes.io/init-containers: '[
```
alanmeadows: Can we transition this to use the new collapsed template:

https://github.com/att-comdev/openstack-helm/blob/master/nova/templates/deployment-api-osapi.yaml#L25

```DEPENDENCY_CONTAINER``` may need to be added to the function in common to support that:

https://github.com/att-comdev/openstack-helm/blob/master/common/templates/snippets/_k8s_init_dep_check.tpl

File: maas/values.yaml [hunk 7]
```diff
@@ -2,9 +2,30 @@
...
+    container: maas-region
+    service:
+    - maas-region-ui
+    - db-service
+
 images:
-  maas_region: quay.io/attcomdev/maas-region:2.1.2-1
-  maas_rack: quay.io/attcomdev/maas-rack:2.1.2-1
+  maas_region: quay.io/attcomdev/maas-region:2.1.2-2
+  maas_toolbox: quay.io/attcomdev/maas-toolbox:2.1.2
```
alanmeadows: Just curious, is a separate toolbox image necessary? A definition for a specific toolbox image is totally reasonable to me--actually having it be unique makes me wonder what it contains that region/rack does not. Just trying to avoid magic images.

maas/templates/bin/_import-boot-resources.sh.tpl
larryrensing: Good catch!

maas/values.yaml
larryrensing: Probably not -- i'll take it off.  maas-region-api is baked into the region image so I can just re-use that one.

maas/templates/bin/_import-boot-resources.sh.tpl
larryrensing: This makes sense-- I noticed with a slower internet connection the job wasn't always completing on the first attempt.  I'll make it configurable for a wider timeout range

maas/templates/job-import.yaml
larryrensing: Done

maas/templates/bin/_maas-region-controller.postinst.tpl
```diff
@@ -62,7 +62,7 @@ if [ "$1" = "configure" ] && [ -z "$2" ]; then
     db_get maas/default-maas-url
     ipaddr="$RET"
     if [ -z "$ipaddr" ]; then
-        ipaddr="maas-region-ui.{{ .Release.Namespace }}"
+        ipaddr="{{ .Values.ui_service_name }}.{{ .Release.Namespace }}"
```
v1k0d3n: nice cleanup, @larryrensing.

maas/templates/bin/_maas-region-controller.postinst.tpl
```diff
@@ -79,8 +79,10 @@ if [ "$1" = "configure" ] && [ -z "$2" ]; then
     # Create the database
     dbc_go maas-region-controller $@
     maas-region local_config_set \
-        --database-host "localhost" --database-name "$dbc_dbname" \
-        --database-user "$dbc_dbuser" --database-pass "$dbc_dbpass"
+        --database-host "localhost" \
+        --database-name "{{ .Values.database.db_name }}" \
+        --database-user "{{ .Values.database.db_user }}" \
+        --database-pass "{{ .Values.database.db_password }}"
```
v1k0d3n: again, clean way to break things out.

larryrensing:
Rebased per changes in https://github.com/att-comdev/openstack-helm/pull/150

common/templates/snippets/_k8s_init_dep_check.tpl
```diff
@@ -8,12 +8,16 @@
   "env": [
     {
       "name": "POD_NAME",
+      {{- if $deps.pod -}}
+      "value": "{{-  $deps.pod . 1  -}}"
```
alanmeadows: Can you walk me through the `. 1` aspect of this?

$deps.pod makes perfect sense

maas/templates/bin/_start.sh.tpl
```diff
@@ -12,3 +18,5 @@ echo 'running postinst'
 
 chmod 755 /var/lib/dpkg/info/maas-region-controller.postinst
 /bin/sh /var/lib/dpkg/info/maas-region-controller.postinst configure
+
+maas-region createadmin --username={{ .Values.credentials.admin_username }} --password={{ .Values.credentials.admin_password }} --email={{ .Values.credentials.admin_email }} || true
```
alanmeadows: This does seem like it would mask a failure to create the region admin - is that an issue?

larryrensing: That depends on how you view the issue at hand, if an admin already exists in postgresql with those credentials (i.e. connecting to persistent storage), maas will throw an error that results in the init container failing to complete.  MaaS supports multiple 'admin' accounts, so redeployment of maas-region with different admin creds will succeed

common/templates/snippets/_k8s_init_dep_check.tpl
larryrensing: fixed

maas/templates/bin/_start.sh.tpl
alanmeadows: Fair enough.  I think the main point is this is as good as we can do it for now.

Can you resolve the conflicts Larry and we will Merge?

alanmeadows:
@larryrensing This was approved but for some reason never got merged.  Can you resolve the conflicts so we can merge it in?

larryrensing:
done @alanmeadows @v1k0d3n

v1k0d3n:
LGTM. @larryrensing thanks for your patience!

PR #151, opened 2017-01-26
make minikube tldr section easier to copy/paste

v1k0d3n:
had a request from community to remove `$` before  each command to make it easier for copy/pasting (and thus a bit more user-friendly). this seemed like a better approach after some thought.

---
This change is [Reviewable](https://reviewable.io/reviews/att-comdev/openstack-helm/151)


v1k0d3n:
trivial merge.

docs/developer/minikube.md
```diff
@@ -53,25 +52,25 @@ $ minikube start \
...
-$ helm init
+helm init
 
 # Package the Openstack-Helm Charts, and push them to your local Helm repository:
-$ make
+make
 
 # Label the Minikube as an Openstack Control Plane node:
-$ kubectl label nodes openstack-control-plane=enabled --all --namespace=openstack
+kubectl label nodes openstack-control-plane=enabled --all --namespace=openstack
```
alanmeadows: We should take this opportunity to include the "openvswitch: enabled" label--which goes on all hosts both tenant and control plane and is a requirement for a functioning neutron and nova boot.  Similarly, we should introduce  "openstack-compute-node: enabled" allowing a host to receive the nova compute-agent and libvirt so that nova VMs can be scheduled there.

v1k0d3n: added the label, but i'm a bit concerned about having non-working ovs labels in minikube docs. i think it will only confuse people (i.e. new users).

v1k0d3n:
@alanmeadows small changes included. i'm going to create an issue about OVS in minikube and mark as a higher priority. if we can figure this out, then I would like to cherry pick the changes into the 0.1.0 tagged release.

PR #154, opened 2017-01-27
Adding postgresql chart

larryrensing:
This change is [Reviewable](https://reviewable.io/reviews/att-comdev/openstack-helm/154)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,12 +1,12 @@
-.PHONY: ceph bootstrap mariadb keystone memcached rabbitmq common openstack neutron nova cinder heat maas all clean
+.PHONY: ceph bootstrap mariadb postgresql keystone memcached rabbitmq common openstack neutron nova cinder heat maas all clean
 
 B64_DIRS := common/secrets
 B64_EXCLUDE := $(wildcard common/secrets/*.b64)
 
-CHARTS := ceph mariadb rabbitmq memcached keystone glance horizon neutron nova cinder heat maas openstack
+CHARTS := ceph mariadb postgresql rabbitmq memcached keystone glance horizon neutron nova cinder heat maas openstack
 COMMON_TPL := common/templates/_globals.tpl
 
-all: common ceph bootstrap mariadb rabbitmq memcached keystone glance horizon neutron nova cinder heat maas openstack
+all: common ceph bootstrap mariadb postgresql rabbitmq memcached keystone glance horizon neutron nova cinder heat maas openstack
 
 common: build-common
 
@@ -16,6 +16,8 @@ ceph: build-ceph
 bootstrap: build-bootstrap
 
 mariadb: build-mariadb
+
+postgresql: build-postgresql
 
 keystone: build-keystone
 
--- a/postgresql/README.md
+++ b/postgresql/README.md
@@ -1,0 +1,11 @@
+# openstack-helm/postgresql
+
+This chart leverages StatefulSets, with persistent storage.
+
+The StatefulSets all leverage PVCs to provide stateful storage to /var/lib/postgresql.
+
+You must ensure that your control nodes that should receive postgresql instances are labeled with openstack-control-plane=enabled, or whatever you have configured in values.yaml for the label configuration:
+
+```
+kubectl label nodes openstack-control-plane=enabled --all
+```
--- a/postgresql/templates/deployment.yaml
+++ b/postgresql/templates/deployment.yaml
@@ -1,0 +1,54 @@
+apiVersion: apps/v1beta1
+kind: StatefulSet
+metadata:
+  name: {{ .Values.service_name }}
+spec:
+  serviceName: {{ .Values.service_name }}
+  replicas: {{ .Values.replicas }}
+  template:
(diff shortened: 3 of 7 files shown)


postgresql/values.yaml
```diff
@@ -0,0 +1,27 @@
...
+
+development:
+  enabled: true
+  storage_path: /data/openstack-helm/postgresql
+
+service_name: postgresql
+
+# using dockerhub postgresql: https://hub.docker.com/r/library/postgres/tags/
+images:
+  postgresql: "postgres:9.5"
```
intlabs: Could we include `docker.io` in the image name - as when running with alternate container runtimes we sometimes need this to be explicit.

postgresql/templates/deployment.yaml
```diff
@@ -0,0 +1,54 @@
+apiVersion: apps/v1beta1
+kind: StatefulSet
+metadata:
+  name: {{ .Values.service_name }}
+spec:
+  serviceName: {{ .Values.service_name }}
+  replicas: 1
```
alanmeadows: This should be in values, with a note that more then one replica is not supported (equivalent to rabbit and memcache today) unless you can determine if there is a way to easily leverage clustering within postgre that is not as much of a herculean effort that mariadb went through.

postgresql/values.yaml
alanmeadows: Agreed.

postgresql/README.md
```diff
@@ -0,0 +1,11 @@
+# aic-helm/postgresql
```
alanmeadows: openstack-helm

Makefile
```diff
@@ -1,12 +1,12 @@
-.PHONY: ceph bootstrap mariadb keystone memcached rabbitmq common openstack neutron nova cinder heat maas all clean
+.PHONY: ceph bootstrap mariadb keystone memcached rabbitmq common openstack neutron nova cinder heat maas postgresql all clean
```
alanmeadows: Just for OCD purposes, lets put this at the front, with the rest of the infrastructure elements, e.g. `ceph bootstrap mariadb postgresql`

postgresql/templates/deployment.yaml
alanmeadows: ... and if this will be our main (and only way) we advocate deploying postgre, we should support clustering.  MaaS doesn't need this to be highly available--its not a problem if it has to wait for it to respawn (as long as we haven't lost the data, which the PVC solves for) but if we use it elsewhere, high availability would likely be a requirement in which case we should solve the clustering challenge.

larryrensing: Understood

postgresql/values.yaml
larryrensing: Good point, done

postgresql/templates/deployment.yaml
```diff
@@ -0,0 +1,54 @@
...
+        resources:
+{{ toYaml .Values.resources | indent 10 }}
+        volumeMounts:
+        - name: postgresql-data
+          mountPath: /var/lib/postgresql
+      volumes:
+{{- if .Values.development.enabled }}
+      - name: postgresql-data
+        hostPath:
+          path: {{ .Values.development.storage_path }}
```
DTadrzak: emptyDir?

larryrensing: I followed a path similar to how mariadb is used in development mode https://github.com/att-comdev/openstack-helm/blob/master/docs/developer/minikube.md

DTadrzak: IMHO mysql should be also changed. With this configuration you are not able to redeploy your OpenStack cluster.

alanmeadows: This is a good point to consider - development mode should be consistent everywhere whatever the direction.

An emptyDir provides the benefit @DTadrzak suggests - its repeatable without the developer needing to remember -- oh I need to clean out that magic directory before repeating.  I am a fan of enabling this because this confuses people -- it confuses me even sometimes.

That said, a hostPath does provide some ability for the developer to potentially do some resiliency testing -- can I restart this container successfully with existing data

On a practical level, hostPath also allows developers to do the needful, i.e. reboot their laptop because we all know we do it, and not lose their running cluster with any test data they injected or setup.

Do we want two development modes to satisfy both cases?

larryrensing: I think that two storage options for development mode would be beneficial.  I'm more keen to lean towards the hostDir approach, simply because it works as a quick solution to persistent storage.  That being said, there have been times where I have forgot that a DB mounted on a hostDir and my deployment starts with a already init'd DB.

larryrensing: @v1k0d3n @wilkers-steve @intlabs

alanmeadows: We've sat on this one for long enough, lets discuss a unified approach toward outfitting an emptyDir/hostDir "path" for each chart as a chart wide design separately so you can enable one way or the other as a developer and let this useful change get in.

PR #156, opened 2017-01-27
BUGFIX: CEPH Mons autodetection doesn't work reliably across install types

galthaus:

I've found that the ceph network discovery doesn't work cleanly for all install types with Kargo (cloud vs bare metal vs calico flannel. ...).  Injecting the MON_IP and public cluster resolve the issue.  We have the information so it seems cleaner to do it this way.

---
This change is [Reviewable](https://reviewable.io/reviews/att-comdev/openstack-helm/156)

### DIFF
--- a/ceph/values.yaml
+++ b/ceph/values.yaml
@@ -25,6 +25,7 @@ labels:
   node_selector_value: enabled
 
 network:
+  public: "10.25.0.0/16"
   port: 
     mon: 6789
     rgw_ingress: 80
--- a/ceph/templates/statefulset-mon.yaml
+++ b/ceph/templates/statefulset-mon.yaml
@@ -72,10 +72,16 @@ spec:
               value: MON
             - name: KV_TYPE
               value: k8s
-            - name: NETWORK_AUTO_DETECT
-              value: "4"
             - name: CLUSTER
               value: ceph
+            - name: NETWORK_AUTO_DETECT
+              value: "0"
+            - name: CEPH_PUBLIC_NETWORK
+              value: {{ .Values.network.public | quote }}
+            - name: MON_IP
+              valueFrom:
+                fieldRef:
+                  fieldPath: status.podIP
           volumeMounts:
             - name: ceph-conf
               mountPath: /etc/ceph
--- a/docs/installation/getting-started.md
+++ b/docs/installation/getting-started.md
@@ -262,7 +262,7 @@ Please ensure that you use ``--purge`` whenever deleting a project.
 ## Ceph Installation and Verification
 Install the first service, which is Ceph. If all instructions have been followed as mentioned above, this installation should go smoothly. Use the following command to install Ceph:
 ```
-admin@kubenode01:~$ helm install --name=ceph local/ceph --namespace=ceph
+admin@kubenode01:~$ helm install --set network.public=$osd_public_network --name=ceph local/ceph --namespace=ceph
 ```
 
 ## Bootstrap Installation


alanmeadows:
This may be a problem, and may directly related to https://github.com/att-comdev/openstack-helm/pull/157

We had to make changes to the upstream image to support autodetection properly.  It would frequently detect IPv6 images, even with "4" specified.

Did you experience this _after_ you switched to the upstream image? This may mean we need to fix your hanging issue rather then switching to the upstream version.

Specifically, fixes here:

https://github.com/att-comdev/dockerfiles/commit/da600c65569c80eb93c6ea6863806a0ed5a43dac

We shouldn't have to specify the network if auto detect is working properly.

galthaus:
Interestingly enough, your upstream change is similar to what I'm doing.  I'm doing it with changing the upstream.

It may have been having issues choosing IPv6 first and that is what was hanging.  It is possible.  The bigger issue that I was having is that the detection through routing is dangerous.  Depending upon what and where you deploy this, the route will not be "sane".   My calico systems end up with routes to 169 addresses.  This doesn't resolve to any address.  I felt it was cleaner to inject the information that we have (or the system has) in this case.  It removes the ambiguity and reliance on upstream.

intlabs:
@alanmeadows we may need to move towards manually specifying the public subnet, though I'd like to hold off as along as possible. It's looking increasingly likely that the ability to autodetect will only become harder moving forward. I've discussed this with a few of the folks in sig-network and it looks like as k8s starts to support more complex topologies though CNI (looking at you Kuryr-Kubernetes, and other fancier implementations) there may not be a single subnet in play - and this will not even be queryable from the API, let alone from within a container :(

alanmeadows:
I can admit that having magic baked into the container to discover is essentially against our philosophy.  I'm good with this update as long as we can get https://github.com/att-comdev/openstack-helm/pull/157 merged first, and pinned to a specific version (as opposed to latest)

v1k0d3n:
i'll second that. @galthaus i think you have a path now. again, thanks for contributing and trying it out with DigitalRebar 👍 

@alanmeadows we need to make note of these conditions in our philosophy documentation ([section 1](https://github.com/att-comdev/openstack-helm/tree/master/docs#1-openstack-helm-design-principals)), so developers are aware of project expectations. also, it's really important that we lay out what we expect for sane and consistent values across charts. in only two months, we have 36 stars on the repo. this tells me that people are interested, and it's really important to address this as i expect contributors to start trickling in (especially when we make the move to either Kubernetes or Openstack repos soon).

alanmeadows:
For sure -- our documentation is in flight and our number one priority to lay these out and begin to reference them in PRs.

v1k0d3n:
@galthaus i cherry picked this change, along with #157, and everything came up as expected. this change is inline with what we really want (using upstream `images:tags`). once you push a `tag` in #157, then i can merge in these  two PR's. i used the `tag-build-master-jewel-ubuntu-16.04` in my tests, but it appears that `latest` is mapped to `tag-build-master-jewel-ubuntu-14.04`, just fwiw.

thanks for these updates! these are extremely helpful changes. :+1:

galthaus:
I'll change latest to tag-build-master-jewel-ubuntu-16.04 in my PR.  Does that sound good?

v1k0d3n:
@galthaus that's perfect.

v1k0d3n:
@alanmeadows and @intlabs let's capture some of the other concerns with ceph for some issues and roadmap. i.e. wrt architectural planning around values structure, etc. i'll mark this issue as a bugfix since it does in fact resolve the current issue (that our ceph deployment doesn't work), and we can plan out and document the framework in more details in issues and between releases.

ceph/values.yaml
```diff
@@ -25,6 +25,7 @@ labels:
   node_selector_value: enabled
 
 network:
+  public: "10.25.0.0/16"
```
galthaus: Yes - It is based upon the getting started doc value apparently from kube_adm.

v1k0d3n: yup that's what i noticed. sounds good. i need to update the minikube doc to outline the same. i can correct this after the merge of this PR.

galthaus:
Is there anything left for me to do?

PR #157, opened 2017-01-27
BUGFIX: Change the ceph images from the custom ATT to the ceph-docker upstream.

galthaus:

The ATT produced ceph image alway seemed to hang for me.  Switching to the ceph provided docker image seemed to work better for me.

---
This change is [Reviewable](https://reviewable.io/reviews/att-comdev/openstack-helm/157)

### DIFF
--- a/ceph/values.yaml
+++ b/ceph/values.yaml
@@ -17,7 +17,7 @@ service:
     name: ceph-mon
 
 images:
-  daemon: ceph/daemon:latest
+  daemon: docker.io/library/ceph/daemon:tag-build-master-jewel-ubuntu-16.04
   pull_policy: IfNotPresent
 
 labels:
--- a/ceph/values.yaml
+++ b/ceph/values.yaml
@@ -17,7 +17,7 @@ service:
     name: ceph-mon
 
 images:
-  daemon: quay.io/attcomdev/ceph-daemon:latest
+  daemon: ceph/daemon:latest
   pull_policy: IfNotPresent
 
 labels:


ceph/values.yaml
```diff
@@ -17,7 +17,7 @@ service:
     name: ceph-mon
 
 images:
-  daemon: quay.io/attcomdev/ceph-daemon:latest
+  daemon: ceph/daemon:latest
```
v1k0d3n: this is an optional change in values which can be modified by the operator (in flight or bring their own values). btw, we really *should not* be using a "latest" tag on our images either. that's a bad practice that can lead to problems in the future.

alanmeadows: I am fine with switching to upstream - as they have incorporated several of our fixes.

That said, agreed -- we have no other chart leveraging :latest -- everything is pinned for a consistent deployment experience.

intlabs: Changing to upstream should be fine, I'm going to try and go over the `entrypoint.sh` script there and make it a bit more sane when I get time. Echoing above, and going slightly further, as well as pinning to a version we should also be specifying that this is an image coming from `docker.io` as when using with other runtimes the shorthand will not always work.

alanmeadows: Agreed, be verbose!

galthaus:
looking to this: docker.io/library/ceph/daemon:tag-build-master-jewel-ubuntu-16.04

v1k0d3n:
agreement between reviewers. merging.

PR #158, opened 2017-01-27
Allow CEPH OSD to use devices on the systems through

galthaus:

On my KVM systems, I have "real" disks attached for ceph consumption.  These changes all the current defaults to be maintained, but real disks can be injected into the configuration.  It assumes system homogeneity but it is a start.

---
This change is [Reviewable](https://reviewable.io/reviews/att-comdev/openstack-helm/158)

### DIFF
--- a/ceph/templates/daemonset-osd.yaml
+++ b/ceph/templates/daemonset-osd.yaml
@@ -60,7 +60,14 @@ spec:
             privileged: true
           env:
             - name: CEPH_DAEMON
-              value: osd_directory
+              value: {{ .Values.osd.daemon }}
+            # Type, Disks, and Zap are ignored for directory-based OSDs.
+            - name: OSD_TYPE
+              value: {{ .Values.osd.type }}
+            - name: OSD_DISKS
+              value: {{ .Values.osd.disks | quote }}
+            - name: OSD_FORCE_ZAP
+              value: {{ .Values.osd.zap | quote }}
             - name: KV_TYPE
               value: k8s
             - name: CLUSTER
--- a/ceph/values.yaml
+++ b/ceph/values.yaml
@@ -30,6 +30,13 @@ network:
     rgw_ingress: 80
     rgw_target: 8088
 
+osd:
+  # Change to osd to use reall disks and set zap to 1
+  daemon: osd_directory
+  type: devices
+  disks: "0:sdb 1:sdc 2:sdd 3:sde"
+  zap: "0"
+
 storage:
   osd_directory: /var/lib/openstack-helm/ceph/osd
   var_directory: /var/lib/openstack-helm/ceph/ceph
--- a/docs/installation/getting-started.md
+++ b/docs/installation/getting-started.md
@@ -265,6 +265,13 @@ Install the first service, which is Ceph. If all instructions have been followed
 admin@kubenode01:~$ helm install --name=ceph local/ceph --namespace=ceph
 ```
 
+By default, ceph will be installed with the system using the space in a host mounted directory.  This can
+be overridden by adding additional set parameters.  This will tell each osd to use physical disks
+/dev/sdb and /dev/sdc as drives.
+```
+admin@kubenode01:~$ helm install --set network.public=$osd_public_network,osd.daemon="osd",osd.zap="1",ods.disks="0:sdb 1:sdc" --name=ceph local/ceph --namespace=ceph
+```
+
 ## Bootstrap Installation
 At this time (and before verification of Ceph) you'll need to install the `bootstrap` chart. The `bootstrap` chart will install secrets for both the `ceph` and `openstack` namespaces for the general StorageClass:
 ```


ceph/templates/daemonset-osd.yaml
```diff
@@ -60,7 +60,14 @@ spec:
             privileged: true
           env:
             - name: CEPH_DAEMON
-              value: osd_directory
```
v1k0d3n: I'd like to see a boolean conditional used for this approach. this allows the  deployment to be more flexible. this is partially our fault...we are in the process of defining where we would want conditionals. docs take a while, but will really help new developers. we need to define still what this ceph conditional would be. let me nail this down with @alanmeadows submit in the PR notes.

Example of conditional in deployment:
example: https://github.com/att-comdev/openstack-helm/blob/master/mariadb/templates/deployment.yaml#L157-L172

docs/installation/getting-started.md
```diff
@@ -265,6 +265,13 @@ Install the first service, which is Ceph. If all instructions have been followed
 admin@kubenode01:~$ helm install --name=ceph local/ceph --namespace=ceph
 ```
 
+By default, ceph will be installed with the system using the space in a host mounted directory.  This can
+be overridden by adding additional set parameters.  This will tell each osd to use physical disks
+/dev/sdb and /dev/sdc as drives.
+```
+admin@kubenode01:~$ helm install --set network.public=$osd_public_network,osd.daemon="osd",osd.zap="1",ods.disks="0:sdb 1:sdc" --name=ceph local/ceph --namespace=ceph
```
v1k0d3n: very nice. i like the override option. ceph is going to be one of those charts we're going to need a lot of conditionals and it will end up very different  than it stands today!

ceph/templates/daemonset-osd.yaml
```diff
@@ -60,7 +60,14 @@ spec:
...
+              value: {{ .Values.osd.daemon }}
+            # Type, Disks, and Zap are ignored for directory-based OSDs.
+            - name: OSD_TYPE
```
alanmeadows: Do any of these need to be conditional if .Values.osd.daemon is osd_directory.  It seems like they should even if just from a clarity perspective--e.g.

```
{{- if not eq .Values.osd.daemon "osd_directory" }}
...
{{- end }}
```

ceph/values.yaml
```diff
@@ -30,6 +30,13 @@ network:
     rgw_ingress: 80
     rgw_target: 8088
 
+osd:
+  # Change to osd to use reall disks and set zap to 1
+  daemon: osd_directory
+  type: devices
+  disks: "0:sdb 1:sdc 2:sdd 3:sde"
```
alanmeadows: We may want to examine this approach more closely -- isn't the ceph best practice recommendation to leverage one osd per physical disk?

Helm can be used to iterate over the disks, creating an OSD per disk.

ceph/templates/daemonset-osd.yaml
galthaus: The entrypoint.sh script doesn't, but it would be a safety thing more than anything else.

ceph/values.yaml
galthaus: That is what the entrypoint script does when you specify the TYPE Devices.  It actually launches and aggregates the logging for each osd.   It is all in the single container.

ceph/templates/daemonset-osd.yaml
galthaus: Not necessarily a problem, but there are more the two ceph osd options, I think.  Now we may want validators that will fail (if that is possible in helm).

v1k0d3n: i agree. this is something we have discuss as well; *"what do we do when there are a, b, c, etc options"*. this is something we  need to lay out in docs for developers who are contributing. we're at that point  now, and we're taking just a little step back  from developing so we can layout clear, sane goals for developers to follow. to this point in our two month project we've written issues and folks have plucked out issues they know we want to collectively solve, but this is changing now that we're trying to find a permanent home for the project (i want these docs drafted now).

ceph/values.yaml
intlabs: I'll need to double-check, but I think it would be better to launch a container per device, rather than rely on the launching of multiple processes that `entrypoint.sh` does. The suggestion Alan has made from his arm-chair would be worth exploring here I think. The primary downside I see here would be the need to specify the OSD/Host topology at chart launch rather than have a somewhat organic approach to deployment, but there I'm on the fence as to whether this constraint is actually a limitation or a sensible belt and braces approach to ensuring our cluster does what we expect.

alanmeadows: The reason for articulating this at the chart level would be two fold -- one is that we do what ceph wants us to do -- one osd process per physical device.  The second is that its very clear what is going to happen at the expense of forcing the operator to be verbose.  I think this is a must when we talk about "zapping" disks or start working through workflows on physical host disk replacement.  Once a container is running multiple processes, you've just given up several kubernetes process management resiliency features, such as restarting the container when the process dies.  I think supervisor and other intermediaries need to have good reason for being in the middle.

I think ending up with chart overrides that look like:

```
osd_map:
  $hostname:
    device_regexp: /dev/sd[e-z]+

or

osd_map:
  $hostname:
    device_list: 
      - /dev/sdb
      - /dev/sdc
...
```

Is not really requiring the operator to go to extreme lengths and it supports varying disk capabilities for each host.

This also is further complicated if operators want to dedicated some devices (e.g. SSDs) to journals as they need to help us identify those.
 
I would also add we should think how we can support /dev/disk-by* type identification/wildcards for this as opposed to spelling out actual device paths such as /dev/sdb.

Even if the gluster use case doesn't require anything like an "osd per device", the mapping of disks to rope in for dedicated storage is likely similar.

ceph/templates/daemonset-osd.yaml
alanmeadows: I'm all for safety when we talk about zapping.

ceph/values.yaml
galthaus: @alanmeadows - I agree, but I was trying to get something to play with quicker to get better feedback on the actual problem space.  I also agree that the way ceph chooses to run multiple OSD processes in the same container misses out on some of the k8s/container value, but it does understand the nuances of running multiple OSD process on the nodes (like port assignment and rotation).  

What you describe is fine and actually works quite nicely with what DR can provide from a hardware enumeration perspective.  DR can build drives for nodes in this environment and update the map as appropriate (even with disk-by-* ids).  My helm foo is not good enough to understand or build this, yet.  I was mostly building this to see what could be done.

So, what do we want to do with this one?

v1k0d3n:
there hasn't been activity on this PR in quite some time. the issue will end up getting dropped once the project is imported in openstack proper (potentially next week).

v1k0d3n:
there doesn't seem to be any interest in contributing back to the requests that have been recommended. closing this PR.

alanmeadows:
Let's make sure to note links to these somewhere in the roadmap document perhaps.  It contains discussion points we want to preserve because even though the issue itself is closed, the work still needs to be done.  The new request, blueprint, whatever form that takes within OpenStack should benefit from this conversation/direction.

v1k0d3n:
@alanmeadows yes, I'm actually in the process of including these items into the roadmap docs and slating the rework for a release (i'm looking at 0.3.0 or 0.4.0 as a target). the discussion points you made about drive zapping, and general drive layout definition in the chart values is a perfectly valid ask/approach for this rework effort.

RichWellum:
I'd be interested in looking at this - looks like a good opportunity to learn for a starter.

Issue #117, opened 2017-01-13
feat: Template Naming Convention


intlabs:
Template names are global, even when used in sub-charts. As a result, it is very easy for collision to occur, with non-deterministic outcomes - which could (and has in my development work) result in effects ranging from failed services to potential security risks.

Due to this, I propose that template names, and macros, should always be scoped to the name of the chart they come from, this would require refactoring, but will I feel make the future development of charts much simpler, and more manageable.

As a result things like:
``` yaml
{{- define "env_ks_user_create_openrc_tpl" }}
```
should be written as:
``` yaml
{{- define "common.env_ks_user_create_openrc_tpl" }}
```
(from: https://github.com/att-comdev/openstack-helm/blob/master/common/templates/snippets/_ks_env_user_create_openrc.tpl#L1)

intlabs:
This has been resolved by merging: https://github.com/att-comdev/openstack-helm/pull/167

PR #167, opened 2017-02-01
Rename common chart, and update all references to functions within it.

intlabs:
This PR renames the common chart to helm-toolkit, and updates all references to components contained within it to be explicit. This prevents the possibility of collision of functions, improves the audibility of charts, and makes it simpler for new contributors to understand where functions come from.

https://github.com/att-comdev/openstack-helm/issues/117

@alanmeadows @wilkers-steve @larryrensing 

---
This change is [Reviewable](https://reviewable.io/reviews/att-comdev/openstack-helm/167)


cinder/templates/_helpers.tpl
```diff
@@ -4,10 +4,10 @@
 
 # this function returns the endpoint uri for a service, it takes an tuple
 # input in the form: service-type, endpoint-class, port-name. eg:
-# { tuple "orchestration" "public" "api" . | include "endpoint_type_lookup_addr" }
+# { tuple "orchestration" "public" "api" . | include "helm_toolkit.endpoint_type_lookup_addr" }
```
alanmeadows: couldn't we use - here for 1:1 with the chart name or does a mix of - for the namespace and _ for variables seem unpalatable?

helm-toolkit/Makefile
```diff
@@ -1,4 +1,4 @@
-EXCLUDE := templates/* charts/* Chart.yaml requirement* values.yaml Makefile utils/* common/Chart.yaml
+EXCLUDE := templates/* charts/* Chart.yaml requirement* values.yaml Makefile utils/* helm-toolkit/Chart.yaml
```
alanmeadows: below do we also want to define secrets as helm(-_)toolkit.$secret in the printf?

Makefile
```diff
@@ -1,14 +1,14 @@
...
+.PHONY: ceph bootstrap mariadb keystone memcached rabbitmq helm-toolkit openstack neutron nova cinder heat maas all clean
 
-B64_DIRS := common/secrets
-B64_EXCLUDE := $(wildcard common/secrets/*.b64)
+B64_DIRS := helm-toolkit/secrets
+B64_EXCLUDE := $(wildcard helm-toolkit/secrets/*.b64)
 
 CHARTS := ceph mariadb rabbitmq memcached keystone glance horizon neutron nova cinder heat maas openstack
-COMMON_TPL := common/templates/_globals.tpl
+COMMON_TPL := helm-toolkit/templates/_globals.tpl
```
alanmeadows: toolkit_tpl?

intlabs: I think `helm-toolkit` is correct here, but will attempt to validate.

helm-toolkit/Makefile
intlabs: Will investigate, I'm currently using my own ceph chart, so will need to (and should...) spin up a cluster using openstack-helm's to develop and test against.

wilkers-steve:
This change looks good -- just a few comments.  Also, looks like some requirements.lock files snuck in.

---

Review status: 0 of 126 files reviewed at latest revision, 3 unresolved discussions.

---

*[Makefile, line 7 at r1](https://reviewable.io:443/reviews/att-comdev/openstack-helm/167#-KbvtZuc_C6GaDTOw6Q--r1-7:-KbyySgnwTZJItDTwFT_:beqa2p0) ([raw file](https://github.com/att-comdev/openstack-helm/blob/d7345a5a5570ae91847497403c9918f0cde99096/Makefile#L7)):*
<details><summary><i>Previously, intlabs (Pete Birley) wrote…</i></summary><blockquote>

I think `helm-toolkit` is correct here, but will attempt to validate.
</blockquote></details>

Agree with Alan.  The var name should TOOLKIT_TPL instead of COMMON_TPL.  Consistency and all that

---

*[cinder/templates/_helpers.tpl, line 7 at r1](https://reviewable.io:443/reviews/att-comdev/openstack-helm/167#-KbvtZuc_C6GaDTOw6Q2-r1-7:-KbyzBOF-_nsJSKiS9dO:b-f4rf7x) ([raw file](https://github.com/att-comdev/openstack-helm/blob/d7345a5a5570ae91847497403c9918f0cde99096/cinder/templates/_helpers.tpl#L7)):*
<details><summary><i>Previously, intlabs (Pete Birley) wrote…</i></summary><blockquote>

I'm not sure, and had a 30 second internal struggle on the issue before just going with '_' as I see valid arguments each way. I have no strong feelings on the subject, so would be happy to change if you see value in using '-', which does have the advantage of making it clearer that this is what we are using to pin a function to a chart.
</blockquote></details>

I think I like '-' here for pinning to the chart name.  It's a little more clear, to me at least.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/att-comdev/openstack-helm/167#-:-Kbz4MSyvxqkEl1yS3OK:bvkpzba)*

Makefile
intlabs: Ah, from @wilkers-steve review just realised I'd missed the point here :) I see it now.

v1k0d3n:
@intlabs this looks good, and I agree the last changes seemed to catch the minor bits that everyone brought up. since it impacts so many files, it will be a bit of a timing issue with merges. we'll pull it in when we clear out some of the backlog of items. getting CI/gating in place will help this a lot going forward.

alanmeadows:
@v1k0d3n With super basic gating in place now (I think it is or is there an outstanding issue?) I see no reason not to let this in.  @intlabs can you conflict resolve? Otherwise, it will continue to drift for @intlabs into conflict land requiring a lot of merge pain.  Let's get this far reaching change in so we can jettison `common`

PR #182, opened 2017-02-07
Add travis-ci to openstack-helm

### LINKED ISSUE #144: ci: basic linting for repository
we need some basic linting and CI gating for our repository. this will make the community more confident in the changes they are submitting, and help assist reviewers before merging code into our repository as well.

wilkers-steve:
Adds the travis.yml file to openstack-helm. The current travis
workflow is: install glide for helm, build helm from source,
run kubeadm in a container, init helm on the client side, then run
tiller locally. The scripts currently run the makefile to
check: helm dep up, helm lint, and helm package on the charts
defined in the makefile. Then a helm install --dry-run debug is
run on each of the charts packaged locally.

**What is the purpose of this pull request?**:  This pull request adds the .travis.yml file to openstack-helm, allowing us to leverage TravisCI for the repository.  It also includes a travis-ci/ directory for housing additional scripts we may want to run as part of the TravisCI jobs.  This will require an administrator to set up TravisCI as an integration on the repository.

**What issue does this pull request address?**: Fixes #144 

**Notes for reviewers to consider**:  The current approach is to leverage the Makefile in the repo for checking: `helm dep up`, `helm lint`, and `helm package`.  We can break these out if necessary.  Also, building and running Tiller from source locally was necessary in order to run jobs that Tiller is required for, such as `helm install --dry-run --debug`.  This approach actually lends us flexibility, because we can then point the locally built Tiller to an external Kubernetes cluster and offload that work instead of trying to run it inside TravisCI.

**Specific reviewers for pull request**: @v1k0d3n @alanmeadows @js5175 @nadeemanwar

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,33 @@
+language: go
+
+services:
+  - docker
+  
+before_install:
+  - export GLIDE_VERSION=v0.12.3
+  - ls $GOPATH/src/
+  - wget "https://github.com/Masterminds/glide/releases/download/$GLIDE_VERSION/glide-$GLIDE_VERSION-linux-amd64.tar.gz"
+  - mkdir -p $HOME/bin
+  - tar -vxz -C $HOME/bin --strip=1 -f glide-$GLIDE_VERSION-linux-amd64.tar.gz
+  - export PATH="$HOME/bin:$PATH" GLIDE_HOME="$HOME/.glide"
+
+install:
+  - cd $GOPATH/src/
+  - mkdir k8s.io && cd k8s.io
+  - git clone https://github.com/kubernetes/helm
+  - cd helm && make bootstrap build
+  - mv bin/helm $HOME/bin
+
+script:
+  - cd $TRAVIS_BUILD_DIR
+  - bash travis-ci/kubeadm_setup.sh  
+  - $HOME/gopath/src/k8s.io/helm/bin/tiller &
+  - export HELM_HOST=localhost:44134
+  - helm init --client-only
+  - helm version
+  - helm serve &
+  - sleep 1m
+  - helm repo add local http://localhost:8879/charts
+  - helm repo update
+  - make
+  - bash travis-ci/charts_dry_run.sh
--- a/travis-ci/charts_dry_run.sh
+++ b/travis-ci/charts_dry_run.sh
@@ -1,0 +1,6 @@
+#!/bin/bash
+
+for chart in *.tgz; do
+  echo "Running helm install --dry-run --debug on $chart";
+  helm install --dry-run --debug local/$chart;
+done
--- a/travis-ci/kubeadm_setup.sh
+++ b/travis-ci/kubeadm_setup.sh
@@ -1,0 +1,5 @@
+#!/bin/bash
+
+docker run -it -e quay.io/attcomdev/kubeadm-ci:latest --name kubeadm-ci --privileged=true -d --net=host --security-opt seccomp:unconfined --cap-add=SYS_ADMIN -v /sys/fs/cgroup:/sys/fs/cgroup:ro -v /var/run/docker.sock:/var/run/docker.sock quay.io/attcomdev/kubeadm-ci:latest /sbin/init
+
+docker exec kubeadm-ci kubeadm.sh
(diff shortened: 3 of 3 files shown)


wilkers-steve:
Current output can be seen here:  https://travis-ci.org/wilkers-steve/openstack-helm

Note: the console output gets truncated due to the length of `helm install --dry-run --debug` output.  We still have access to the raw logs though, which will give a .txt-ish output log for individuals to view.

.travis.yml
```diff
@@ -0,0 +1,33 @@
...
+before_install:
+  - export GLIDE_VERSION=v0.12.3
+  - ls $GOPATH/src/
+  - wget "https://github.com/Masterminds/glide/releases/download/$GLIDE_VERSION/glide-$GLIDE_VERSION-linux-amd64.tar.gz"
+  - mkdir -p $HOME/bin
+  - tar -vxz -C $HOME/bin --strip=1 -f glide-$GLIDE_VERSION-linux-amd64.tar.gz
+  - export PATH="$HOME/bin:$PATH" GLIDE_HOME="$HOME/.glide"
+
+install:
+  - cd $GOPATH/src/
```
alanmeadows: By building helm each time, won't we be adding substantial build time to the checks? Last time I built helm (2.0.0 era) it took a bit.

Also, this would leverage master, and thus any issues introduced to help there (master is not guaranteed to be stable) could effect our CI/CD.  Shouldn't this be pinned and updated as our charts support newer versions of helm?

wilkers-steve:
@v1k0d3n sounds good.  I just filed an issue for getting a full `helm install foo` job set up in TravisCI and will continue to work against that this evening.

travis-ci/charts_dry_run.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+
+for chart in *.tgz; do
+  echo "Running helm install --dry-run --debug on $chart";
+  helm install --dry-run --debug local/$chart;
```
alanmeadows: If there is a kubernetes (albeit inside docker) available, are we planning to do non-dry-run installs in a later phase (at least of those components that makes sense, e.g. not ceph or maas)

wilkers-steve: @alanmeadows Yeah.  That's the logical next step.  I've started pursuing that this afternoon, and have filed issue #183 to cover that point.

.travis.yml
wilkers-steve: It's not really substantial here, since we're building it once then running all of our checks for a particular PR against that Helm build.  The current time for running the entire setup is roughly 7 minutes.

Yeah, the master bit is a fair point.  I agree with pinning this and updating versions as necessary.

alanmeadows: Presumably its to get a tiller binary, I would say an improvement is just  docker run  gcr.io/kubernetes-helm/tiller for expediency (55K image) - plus its guaranteed to be "stable" more so then the build process

wilkers-steve: Yeah, that would definitely be an improvement.  I didn't think to look to see if there was a binary to run Tiller locally instead of in-cluster.  Once again, seems I ignored the mantra:  everything's going to run in a container. 😉

v1k0d3n: @wilkers-steve this can be done on the container-side; pretty easy fix, and reasonable ask.

@alanmeadows this will always leverage master *as a PR is created*. this actually *prevents* issues from appearing on master which is why it's so critical to get in place.

wilkers-steve: @v1k0d3n I think @alanmeadows was referring to targeting Helm master each time it's built, which could introduce some risk in the CI/CD pipeline.

v1k0d3n: @wilkers-steve oh well yes...in that case definitely. #number 1 rule for developers: don't rely on master, and don't use latest tags on containers unless you *want* to introduce problems.

@alanmeadows sorry for the misunderstanding.

wilkers-steve: @v1k0d3n Actually, I'm glad you brought that up..

v1k0d3n:
@wilkers-steve as long as someone commits to the change (anyone can commit to it), then i would like to pull in these changes to start the github parts that are needed.

wilkers-steve:
@v1k0d3n go for it.  I can add whatever tweaks are necessary after it's been pulled in.  One being that kubeadm-ci image needs to be retagged after the changes I submitted last night.  It's currently running off the `:latest` tag.

v1k0d3n:
ok, you got it. i'll create issues for the remaining points brought up.

PR #206, opened 2017-02-15
Replace Ansible with SQLAlchemy/PyMySQL for database creation.

intlabs:
**What is the purpose of this pull request?**: To replace ansible with PyMySQL for DB Creation

**What issue does this pull request address?**: This is part of the effort to become image agnostic.

**Notes for reviewers to consider**: This PR is intended to supersede https://github.com/att-comdev/openstack-helm/pull/140

**Specific reviewers for pull request**: @alanmeadows, @srwilkers, @v1k0d3n


helm-toolkit/templates/snippets/_db_secret_root.yaml.tpl
```diff
@@ -0,0 +1,12 @@
+{{- define "helm-toolkit.secret_db_root" }}
```
alanmeadows: This is not named after the file that contains it.

keystone/templates/secret-db-root.env.yaml
```diff
@@ -0,0 +1,2 @@
+{{- $envAll := . }}
+{{- include "helm-toolkit.secret_db_root" $envAll }}
```
alanmeadows: Because this toolkit script leverages only values from `.Values.database` I would recommend that is what we pass.  We have a mix between passing macros `$envAll` or `.` and the leafs that contain the data they need.  We should standardize on one.  I prefer leafs (`.Values.database` as its clear what input we are passing the macro but I'm flexible.

keystone/values.yaml
```diff
@@ -51,6 +51,9 @@ network:
   ip_address: "0.0.0.0"
 
 database:
+  secret:
```
alanmeadows: It's not clear to me how this differs from the `root_password` below

heat/templates/job-db-init.yaml
```diff
@@ -29,17 +29,37 @@ spec:
...
           {{- end }}
           env:
-          - name: ANSIBLE_LIBRARY
-            value: /usr/share/ansible/
+            - name: ROOT_DB_CONNECTION
+              valueFrom:
+                secretKeyRef:
+                  name: {{ .Values.database.secret.root }}
+                  key: DB_CONNECTION
+            - name: OPENSTACK_CONFIG_FILE
```
alanmeadows: Seems like this will keep growing, maybe our checker-helpers needs its own oslo config file which is just a config map? This may grow to unwieldy proportions.  Understandably could be future work.

keystone/templates/job-db-init.yaml
```diff
@@ -19,23 +19,47 @@ spec:
...
+            - /tmp/db-create.py
           volumeMounts:
-            - name: keystone-bin
-              mountPath: /tmp/init.sh
-              subPath: init.sh
+            - name: configmap-bin
+              mountPath: /tmp/db-create.py
+              subPath: db-create.py
+              readOnly: true
+            - name: pod-etc-service
```
alanmeadows: Just a minor nit, not sure I fully grasp the naming conventions here... pod-etc-service really is an emptyDir, service-config is keystone.conf... I had to look down below two or three times - which either represents my own senility or the need to make this clearer

keystone/templates/secret-db-root.env.yaml
intlabs: That approach works for me.

keystone/values.yaml
intlabs: This is the name of the secret, rather than the actual password. I'll try to make this clearer in a follow-up commit.

alanmeadows: That makes _much_ more sense.  I think just secret_name vs secret would of made me realize this.

keystone/templates/job-db-init.yaml
intlabs: I've used the naming conventions from harbor here:
 - Prefixes:
   - pod: represents a volume that is local to the pod, almost always an emptyDir
   - service: a volume that is common the service, almost always a configMap, but may be a shared volume
 - Suffixes:
   - The mount location of the volume if only used once.
   - If a configmap, just config (unless several are used, in which case I added a further suffix with some brief indicator of the contents)

intlabs:
To resume work on this PR should merge the following:
 - https://github.com/att-comdev/openstack-helm/pull/275
 - https://github.com/att-comdev/openstack-helm/pull/281
 - https://github.com/att-comdev/openstack-helm/pull/273
As they make changes to the job layout that will cause conflicts with this work.

keystone/templates/secret-db-root.env.yaml
intlabs: This work (and comments) have been made moot by: https://github.com/att-comdev/openstack-helm/pull/262/files#diff-728a997bda2af0665cbf1fde10e4231fR153, which is a better approach. For now I'll adopt this method verbatim, to allow work to continue on this PR.

keystone/values.yaml
intlabs: This will be refactored to match the style here: https://github.com/alanmeadows/aic-helm/blob/b40a7bf3be5bcc9960452b0ad72c6db9932f101e/keystone/values.yaml#L303. I will insert the key `k8s_secret_name` as appropriate within this to define the name of the secret that will hold the derived SQLAlchemy connection string.

heat/templates/job-db-init.yaml
intlabs: The idea here is to consume the same config file as the service will consume, the job itself is what defines the scope here I think. Unless we are across purposes?

intlabs:
Superseded by https://github.com/att-comdev/openstack-helm/pull/289

Issue #187, opened 2017-02-08
ci: travis-ci dry-run optimizations


alanmeadows:
1) Suggest dry-run logic is put into make so developers can run it locally:

`make dry-run`

2) Suggest dry-run logic pipes output: `helm .... 2>&1 > /tmp/dry-run-output.log` for each chart and simply validates that `helm --dry-run` exited with 0.  If it is !=0 then it does a `cat /tmp/dry-run-output.log`.  This solves out huge output issue as charts rendering fine are not of interest.

3) We should do a grep on `/tmp/dry-run-output.log` to look for `'gotpl'` - this surfaces when there is a pipeline issue but does not result in a `helm ... --dry-run` error or exit status of non-zero.  To test, edit a pipeline, such as the `init-dep-check` in a random chart, change it to a non-existent value, and run it through dry-run.  You will see lines in the output such as `template: .... 'gotpl' ... error` but it will not fail the dry-run.

wilkers-steve:
@alanmeadows Sounds good.  These changes shouldn't be difficult, and I can work on getting these in over the next day or two

nadeemanwar:
@alanmeadows what you do when you found the error like 'gotpl', fail the build or just print it in the log?

Running helm install --dry-run --debug on maas-0.1.0.tgz
2017/02/16 03:03:06 storage.go:37: Getting release "boisterous-chipmunk" (v1) from storage
2017/02/16 03:03:06 release_server.go:826: Dry run for boisterous-chipmunk
No Error found for dry run checking for gotpl
    template: no template "helm_toolkit.template" associated with template "gotpl"
    template: no template "dep_check_init_cont" associated with template "gotpl"

alanmeadows:
This is a failed build when this occurs.  If we find this, there is no way the template will have been rendered correctly.

nadeemanwar:
@alanmeadows  I updated the script to fail the build when 'gotpl' is found the log. Now the build is failing on this one. How to fix it :-)  This is the last log from my testing which is to fix the part 2 and 3 of the issue.
https://travis-ci.org/nadeemanwar/openstack-helm/builds/202304181

nadeemanwar:
I added a check to see certain phrases in the output log because even when the dry-run install is successful we can still see errors in the log. I added a erros.list file in repo where we can define the phrases which we should search in the output log after the installation of chart and if we find something, fail the build.

[...]

Issue #230, opened 2017-02-21
bug: Missing template for nova configmaps


wilkers-steve:
**Is this a bug report or feature request?** (choose one): Bug report

**Kubernetes Version** (output of `kubectl version`): N/A

**Helm Client and Tiller Versions** (output of `helm version`): N/A

**Development or Deployment Environment?**: N/A

**Release Tag or Master**: Master

**Expected Behavior**: Nova chart lints and installs with `--dry-run` successfully

**What Actually Happened**:  Error reported during `helm install --dry-run` due to missing template

**How to Reproduce the Issue** (as minimally as possible):  Run `helm install nova --dry-run --debug`

**Any Additional Comments**:

This issue is raised because of the template declared in the nova etc configmap found here: https://github.com/att-comdev/openstack-helm/blob/master/nova/templates/configmap-etc.yaml#L22.  We can look in the /etc directory in nova and see there's a template for _ceph.conf.tpl.  I'm assuming this was just a misnaming here.

gardlt:
it seem both _ceph.conf.tpl and _ceph.client.cinder.keyring.yaml.tpl are needed. we might be missing a tpl or having a bad import.

intlabs:
@wilkers-steve, @gardlt I think I need to that the hold my hands up to this one, looks like I butchered this as part of my cleanup work. See here, for the lines I accidently removed: https://github.com/att-comdev/openstack-helm/blob/8afa729ffba9378b76e63dfc282f16ecf1d9baa9/nova/templates/configmap-etc.yaml#L7-L8.
I'll put a fix in asap.

PR #231, opened 2017-02-21
WIP - Limiting the log printing

nadeemanwar:
#187 

- Try to catch the errors or warning which are not caught in the dry-run
- Limit the log which is getting printed in travis-ci log

  Issue no.  187

- This is still work in progress. Asking for some feedback
- Errors and warnings phrases are read from the which can be modified without changing the code
- Errors and warnings are printed at the bottom of the log in the collapsible section for easy access
- Link to the last run for reference.  https://travis-ci.org/nadeemanwar/openstack-helm/builds/203919986


travis-ci/errors.list
```diff
@@ -0,0 +1,2 @@
+no template "nova/templates/etc/_ceph.client.cinder.keyring.yaml.tpl" associated with template "gotpl"
```
wilkers-steve: I'm assuming these are used for testing.  We won't need exact phrases targeting specific charts and templates, but it seems this is working as a test case.

travis-ci/charts_dry_run.sh
```diff
@@ -14,7 +14,35 @@
...
+    cat travis-ci/errors.list
+    if [ `egrep -f travis-ci/errors.list /tmp/dry-run-output.log | wc -l` -ne 0 ]; 
+    then
+      echo "Found errors, setting the status to 1 and printing the log" 
+      status=1;
+      egrep -f travis-ci/errors.list /tmp/dry-run-output.log
+      cat /tmp/dry-run-output.log >> /tmp/dry-run-errors.log
+    fi
+    echo "No Error found for dry run checking for for these warning phrases"
+    cat travis-ci/warnings.list
```
wilkers-steve: Similar here.

travis-ci/charts_dry_run.sh
```diff
@@ -14,7 +14,35 @@
...
-  helm install --dry-run --debug local/$chart;
+  helm install --dry-run --debug local/$chart 2>&1 > /tmp/dry-run-output.log
+  if [ $? -ne 0 ];
+  then
+    echo "Found error and setting status to 1"
+    cat /tmp/dry-run-output.log >> /tmp/dry-run-errors.log
+    status=1;
+  else
+    echo "No Error found for dry run checking for these error phrases"
+    cat travis-ci/errors.list
```
wilkers-steve: We may not need to advertise what errors we're checking for specifically.  We can just print out any we come across that match the words and phrases we'd like to search against.  This helps us prune the console output a little more.

travis-ci/charts_dry_run.sh
```diff
@@ -14,7 +14,35 @@
...
   echo "Running helm install --dry-run --debug on $chart";
-  helm install --dry-run --debug local/$chart;
+  helm install --dry-run --debug local/$chart 2>&1 > /tmp/dry-run-output.log
+  if [ $? -ne 0 ];
+  then
+    echo "Found error and setting status to 1"
+    cat /tmp/dry-run-output.log >> /tmp/dry-run-errors.log
+    status=1;
+  else
+    echo "No Error found for dry run checking for these error phrases"
```
wilkers-steve: We should make this more verbose.  At first glance, this appears that there's no errors found for dry run checking but we're looking for error phrases.  This would be more readable if it were along the lines of "Install successful.  Checking for template issues" or something similar

travis-ci/charts_dry_run.sh
```diff
@@ -14,7 +14,35 @@
...
+    cat travis-ci/errors.list
+    if [ `egrep -f travis-ci/errors.list /tmp/dry-run-output.log | wc -l` -ne 0 ]; 
+    then
+      echo "Found errors, setting the status to 1 and printing the log" 
+      status=1;
+      egrep -f travis-ci/errors.list /tmp/dry-run-output.log
+      cat /tmp/dry-run-output.log >> /tmp/dry-run-errors.log
```
wilkers-steve: Instead of jamming the entire output into the error log, could we only append the line that matches the previous grep?  This cuts down on the error messages being reported back to the console and makes it easy to determine what needs fixing directly instead of combing through all of the output for that chart.

nadeemanwar: Do you think one line will be enough to debug it. I was thinking of just printing one line in the main log above to show what is the error but then print the full log for only those charts which failed at the end if they want to see the detail

travis-ci/charts_dry_run.sh
nadeemanwar: Agree. This is just for debugging now.

travis-ci/charts_dry_run.sh
wilkers-steve: Well, let's look at the output and find out.  For some of the issues related to `helm install foo ---dry-run --debug`, it's really just one line explaining what caused the error and what file.  That could be enough for us to see what needs changing.  The problem with printing the whole log is that usually it's a few lines hiding in a wall of text rendered from the Helm templates.  Someone could surely just `ctrl+f` to find those keywords, but really we want to be as surgical as possible in the console output.

travis-ci/charts_dry_run.sh
wilkers-steve: Awesome.  Sounds good 👍

travis-ci/charts_dry_run.sh
js5175: Do we want to count and rollup: # of successfully installed/verified charts, chart warnings, and failures?

wilkers-steve: I think it'd help to know which charts installed successfully and which did not, and format them in a human readable fashion

nadeemanwar: I made that change it is now printing the summary

travis-ci/charts_dry_run.sh
nadeemanwar: commented that out

travis-ci/charts_dry_run.sh
nadeemanwar: Make the change to just print out the grep lines

travis-ci/charts_dry_run.sh
nadeemanwar: made that change too.

js5175:
So the nova chart is the failing:
template: no template "nova/templates/etc/_ceph.client.cinder.keyring.yaml.tpl" associated with template "gotpl"

nadeemanwar:
This is what Steven wrote about the nova issue in the issue 187 comment log.  "The nova issue is related to a template declared for the nova etc configmap. I've filed an issue here: #230. good catch"

alanmeadows:
I agree warning mode is fine until the gotpl issues are fixed that we already have.

Can we apply the same silence-or-show-only-if-command fails on the helm setup portions?

The actual testing of our code (given we are not actually looking to test the helm installation process itself) doesn't begin until 1000 lines into the results.  I am trying to make it clear what I developers are interested in seeing.

https://travis-ci.org/att-comdev/openstack-helm/builds/204212220#L953

Alan Meadows

travis-ci/errors.list
alanmeadows: Confused about this as well, I thought I would something generic here like `associated with template "gotpl"`

nadeemanwar:
should we put a flag in the yaml file like "show-all-log" so if that is set true, if should print the log otherwise just skip the log if there are not errors or just silence all the log except the errors or status?

travis-ci/errors.list
nadeemanwar: Yes it need to be generic one but I don't know exactly what to put. I put "associated with template "gotpl"" before but it look like the problem we got for nova has the same wording as the one which is in the cinder.  As per Steven, the cinder one is not exactly an issue, so I moved it to warning.  So if we put "associated with template "gotpl"", is this a warning or error?

Issue #243, opened 2017-03-03
mariadb chart fails with kolla/centos-binary-mariadb image


wilreichert:
**Is this a bug report or feature request?** (choose one):
bug 

**Kubernetes Version** (output of `kubectl version`):
Client Version: version.Info{Major:"1", Minor:"5", GitVersion:"v1.5.3", GitCommit:"029c3a408176b55c30846f0faedf56aae5992e9b", GitTreeState:"clean", BuildDate:"2017-02-15T06:40:50Z", GoVersion:"go1.7.4", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"5", GitVersion:"v1.5.3", GitCommit:"029c3a408176b55c30846f0faedf56aae5992e9b", GitTreeState:"clean", BuildDate:"2017-02-15T06:34:56Z", GoVersion:"go1.7.4", Compiler:"gc", Platform:"linux/amd64"}

**Helm Client and Tiller Versions** (output of `helm version`):
Client: &version.Version{SemVer:"v2.2.1", GitCommit:"db531fd75fb2a1fb0841a98d9e55c58c21f70f4c", GitTreeState:"clean"}
Server: &version.Version{SemVer:"v2.2.1", GitCommit:"db531fd75fb2a1fb0841a98d9e55c58c21f70f4c", GitTreeState:"clean"}

**Development or Deployment Environment?**:
deployment 3 node cluster w/ ceph 

**Release Tag or Master**:
master

Steps to reproduce:
```
% git clone clone https://github.com/att-comdev/openstack-helm
<<<cut>>>
% cd openstack-helm
% helm install mariadb --set images.mariadb="kolla/centos-binary-mariadb:master" --set volume.class_name=ceph --namespace test --name testdb
<<<cut>>>
<<<wait a couple minutes >>>
% kubectl get po -n test
NAME                 READY     STATUS              RESTARTS   AGE
mariadb-0            0/1       Pending             0          1m
mariadb-seed-7mqq8   0/1       ContainerCreating   0          2s
% kubectl logs mariadb-seed-7mqq8 -n test
INFO:__main__:Loading config file at /var/lib/kolla/config_files/config.json
ERROR:__main__:Unexpected error:
Traceback (most recent call last):
  File "/usr/local/bin/kolla_set_configs", line 412, in <module>
    main()
  File "/usr/local/bin/kolla_set_configs", line 401, in main
    config = load_config()
[...]

PR #244, opened 2017-03-03
use pod command: instead of the COMMAND environment variable

wilreichert:
addresses #243


v1k0d3n:
i'm ok with this change. it brings mariadb more in line with keystone, however we know that mariadb/keystone still need some rework. thanks for the PR @wilreichert 

adding @intlabs and @alanmeadows as reviewers, just in case they want to raise any issues i'm overlooking.

intlabs:
LGTM, I think with this we are now setting the entry point in a unified manner for all pods across all charts :)

v1k0d3n:
good catch, and thanks for the PR @wilreichert!

PR #247, opened 2017-03-05
docs: general updates to documentation

v1k0d3n:
General Documentation Cleanup:
* Updating documentation for the inclusion of `etcd-rabbitmq` chart
* Moved troubleshooting items in getting-started to troubleshooting section
* Update documented versions for Kubernetes (1.5.2), Helm (2.2.1), etc.
* Minor other document fixes


wilkers-steve:
This looks good.  Thanks for the cleanup here.  Let me know if you're looking at add any further commits.  This looks good to me as is with what you've included.

LGTM

docs/developer/minikube.md
```diff
@@ -64,6 +64,7 @@ kubectl label nodes openstack-control-plane=enabled --all --namespace=openstack
 # Deploy each chart:
 helm install --name mariadb --set development.enabled=true local/mariadb --namespace=openstack
 helm install --name=memcached local/memcached --namespace=openstack
+helm install --name=etcd-rabbitmq local/etcd --namespace=openstack
```
intlabs: This chart is not present in the tag `0.1.0`. We should add a comment here to indicate this, as errors here will probably confuse new users/developers.

v1k0d3n: good point. i think what may be even better is to just convert everything to a 0.2.0, properly branch this time, and then reference 0.2.0 from this point forward. this way we can drop the 0.1.0 tag sprawl/issues.

v1k0d3n: actually, the more that i think about this...removing the specific 0.1.0 branch references would be best, and instruct users to use the "latest stable branch", which in this case will be 0.2.0 (the first true branch). will that work pete? this is a small detail, but helpful for users. i've noticed that a majority of users are using master regardless of our docs referencing 0.1.0. true branches will help going forward.

v1k0d3n:
@alanmeadows can you review/merge this in please (unless you have issues with it)?

docs/developer/minikube.md
v1k0d3n: @intlabs can you let me know if this is acceptable or if you were expecting something else? this week i'd like to cut a legitimate 0.2.0 branch (most likely from commit: [6e0cda0](https://github.com/att-comdev/openstack-helm/tree/6e0cda0f52ff61e5cfd7e8ada0ea8c5fe107f89d), test and reference that for docs. i'd also like to cut a release of 0.1.0 (again as a branch), cherry pick the commits for Ceph and have that be an "official" first release. at that point, we'll have some pretty good places to start for releases. release of 0.3.0 would be towards the end of the month when the milestone is complete. and that's what i foresee ending up in openstack proper imo (or unless someone completely objects).

intlabs: @v1k0d3n sorry for the lag in responding. Removing references to 0.1.0 and moving to 'latest stable branch' is a great idea, so +2 from me.

PR #256, opened 2017-03-08
Converted readiness.py to bash script.

### LINKED ISSUE #253: mariadb chart readiness probe fails with kolla/*-binary-mariadb image
The mariadb chart [readiness probe](https://github.com/att-comdev/openstack-helm/blob/master/mariadb/templates/bin/_readiness.py.tpl#L19) uses pymysql which is not installed in the kolla mariadb images. End result is the database starts fine but the container is never marked as read due to the probe failure.

andrecolin:
**What is the purpose of this pull request?**:  Replace readiness python script since python not in kolla base image

**What issue does this pull request address?**: Fixes  #253

**Notes for reviewers to consider**: Still need to convert peerfinder python script next


wilreichert:
pulled your patch & tested on our cluster with 
```
helm install --name mariadbtest \
--set images.mariadb=kolla/ubuntu-binary-mariadb:3.0.2 \
--set replicas=1 \
--set volume.class_name=ceph \
local/mariadb --namespace=test
```

Looks like you missed a bash in the deployment.yaml

```
% kubectl get ev -n test | grep Readiness                                                                                                                 :( 1 17-03-09 - 4:23:01
6s         3m          21        mariadb-0   Pod       spec.containers{mariadb}   Warning   Unhealthy          {kubelet oreo05}       Readiness probe failed: rpc error: code = 2 desc = oci runtime error: exec failed: container_linux.go:247: starting container process caused "exec: \"/tmp/mariadb-readiness.sh\": permission denied"
```

andrecolin:
@wilreichert Can you pull again and test on your server

wilreichert:
Single instance case above works fine now but clustered still has issues. When launching

```
helm install --name mariadbtest \
--set images.mariadb=kolla/ubuntu-binary-mariadb:3.0.2 \
--set replicas=3 \
--set volume.class_name=ceph \
local/mariadb --namespace=test
```

it looks like the peer-finder script has perms isues & init goes bad - [pod logs](https://gist.github.com/wilreichert/28e12cc54472052b5d0ad16daab17fc1)

FWIW, I made sure to nuke pvc's in-between tests.

andrecolin:
@wilreichert There were some extra items in the script bang entry. Can you try again

wilreichert:
hmm, same problem - https://gist.github.com/wilreichert/df413f6ab6ae918bfb747fa6d061978c

mariadb/templates/bin/_seed.sh.tpl
```diff
@@ -49,7 +49,7 @@ function wait_for_cluster {
             continue
         fi
         # Count number of endpoint separators.
-        ENDPOINTS_CNT=`python /tmp/peer-finder.py mariadb 1 | grep -o ',' | wc -l`
+        ENDPOINTS_CNT=`/tmp/peer-finder.sh mariadb 1 | grep -o ',' | wc -l`
```
intlabs: this line should read: ``ENDPOINTS_CNT=`bash /tmp/peer-finder.sh mariadb 1 | grep -o ',' | wc -l``
which will fix the permissions issue.

andrecolin:
Found two places that did an eval and bash was missing from the eval. Permission issue should be gone

andrecolin:
Redoing pull request

PR #261, opened 2017-03-09
Enable nodeport on/off for services

### LINKED ISSUE #257: feat: NodePort on/off support
**Is this a bug report or feature request?** (choose one):
Feature request

**Kubernetes Version** (output of `kubectl version`):
Any
**Helm Client and Tiller Versions** (output of `helm version`):
Any
**Development or Deployment Environment?**:
Both
**Release Tag or Master**:
Master
**Expected Behavior**:
Services should have a flag to enable/disable node ports
**What Actually Happened**:
Only Horizon has this currently
**How to Reproduce the Issue** (as minimally as possible):
n/a
**Any Additional Comments**:
Node ports need to have a consistent numerical pattern across charts, default kubernetes NodePort range is from 30000-32767.

larryrensing:
Provide boolean to enable/disable nodeports for services
**What issue does this pull request address?**: Fixes #257 

Heat values.yaml was formatted differently than the others, did a small modification to keep a consistent naming scheme for ports. 

@v1k0d3n @wilkers-steve @alanmeadows @intlabs

### DIFF
--- a/cinder/values.yaml
+++ b/cinder/values.yaml
@@ -62,8 +62,10 @@ keystone:
   cinder_region_name: "RegionOne"
 
 network:
+  enable_node_port: false
   port:
     api: 8776
+    node_port_api: 30776
 
 database:
   address: mariadb
--- a/glance/templates/service-api.yaml
+++ b/glance/templates/service-api.yaml
@@ -19,5 +19,11 @@ metadata:
 spec:
   ports:
     - port: {{ .Values.network.port.api }}
+    {{ if .Values.network.enable_node_port }}
+      nodePort: {{ .Values.network.port.node_port_api }}
+    {{ end }}
   selector:
     app: glance-api
+  {{ if .Values.network.enable_node_port }}
+  type: NodePort
+  {{ end }}
--- a/heat/templates/etc/_heat.conf.tpl
+++ b/heat/templates/etc/_heat.conf.tpl
@@ -22,9 +22,9 @@ deferred_auth_method = "trusts"
 enable_stack_adopt = "True"
 enable_stack_abandon = "True"
 
-heat_metadata_server_url = {{ .Values.service.cfn.proto }}://{{ .Values.service.cfn.name }}:{{ .Values.service.cfn.port }}
-heat_waitcondition_server_url = {{ .Values.service.cfn.proto }}://{{ .Values.service.cfn.name }}:{{ .Values.service.cfn.port }}/v1/waitcondition
-heat_watch_server_url = {{ .Values.service.cloudwatch.proto }}://{{ .Values.service.cloudwatch.name }}:{{ .Values.service.cloudwatch.port }}
+heat_metadata_server_url = {{ .Values.service.cfn.proto }}://{{ .Values.service.cfn.name }}:{{ .Values.network.port.cfn }}
+heat_waitcondition_server_url = {{ .Values.service.cfn.proto }}://{{ .Values.service.cfn.name }}:{{ .Values.network.port.cfn }}/v1/waitcondition
+heat_watch_server_url = {{ .Values.service.cloudwatch.proto }}://{{ .Values.service.cloudwatch.name }}:{{ .Values.network.port.cloudwatch }}
 
 num_engine_workers = {{ .Values.resources.engine.workers }}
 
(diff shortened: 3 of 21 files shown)


cinder/values.yaml
```diff
@@ -62,8 +62,10 @@ keystone:
   cinder_region_name: "RegionOne"
 
 network:
```
wilkers-steve: In order to keep the node port information grouped together, it'd read better to have a `node_port` section under `network`, then scope the `enabled` flag and the `node_port_api` port entries underneath that section.  What are your thoughts?

glance/templates/service-api.yaml
```diff
@@ -19,5 +19,11 @@ metadata:
 spec:
   ports:
     - port: {{ .Values.network.port.api }}
+    {{ if .Values.network.enable_node_port }}
```
wilkers-steve: With the change to values.yaml, this would read `if .Values.network.node_port.enabled`, which seems more clear and aligns with how we handle booleans in other sections, like development mode

heat/templates/etc/_heat.conf.tpl
```diff
@@ -22,9 +22,9 @@ deferred_auth_method = "trusts"
 enable_stack_adopt = "True"
 enable_stack_abandon = "True"
 
-heat_metadata_server_url = {{ .Values.service.cfn.proto }}://{{ .Values.service.cfn.name }}:{{ .Values.service.cfn.port }}
```
alanmeadows: I think the separation of the service vs container binding was actually useful.  @intlabs please weigh in.

This may become moot because I would hope pete's endpoint resolution stuff likely replaces lines such as these.

This raises a good question of whether enabling endpoints actually alters standard endpoint logic (building these from `endpoints:`) or its just something exposed so humans can talk to them.

cinder/values.yaml
alanmeadows: It looks like my comment got lost here in a shuffle this morning.

I was going to take a twist on Steve's feedback.

Essentially since this is a service level modification, taking heat as an example:

```
service:
  api:
    name: "heat-api"
    port: 8004
    proto: "http"
  cfn:
    name: "heat-cfn"
    port: 8000
    proto: "http"
  cloudwatch:
    name: "heat-cloudwatch"
    port: 8003
    proto: "http"
```

becomes:

```
service:
  api:
    name: "heat-api"
    port: 8004
    proto: "http"
    node_port:
      enabled: false
      port: 123456
  cfn:
    name: "heat-cfn"
    port: 8000
    proto: "http"
    node_port:
      enabled: false
      port: 123456
  cloudwatch:
    name: "heat-cloudwatch"
    port: 8003
    proto: "http"
    node_port:
      enabled: false
      port: 123456
```

To me this has some value as I can enable node ports on just select services.  I can imagine a possible use case where this is used beyond just development environments, so someone may want to get very picky about what particular service within a chart they want to expose in this fashion.

v1k0d3n:
@larryrensing i merged in the `service` labels for `network` earlier today, and that caused your merge conflicts. simple changes, but sorry for the re-work. at least it's small/limited.

i agree with @alanmeadows' approach here; per service/network declaration. the example is great, and this provides greater flexibility for operators. 

it may be a bit late for this feedback as it pertains to GitHub...but something that could be useful going forward even in gerrit is to pick a single service-level chart as an example, keep the work small, ask for feedback, and then expand on the concepts once you receive initial feedback on the PR or patchset. this will keep the work laser-focused/targeted, easier for initial review of the introduced concepts, reduce your required *rework*, open up the feedback loop for input earlier (before you've invested a large chunk of time), and then the re-work can be applied as a more sweeping follow-up change in the future. 

i just a thought that could help you, @wilkers-steve or anyone else as they're developing and getting initial feedback from @alanmeadows or other community members on conceptual or architectural changes. this is something that i'd like to see not just from our team going forward, but it's a practice i'd like to see from the general community as well. i think it will help limit the introduction of large or disruptive architectural shifts in strategy. as an example, look at @alanmeadows targeted architectural changes/concepts for configuration (#262). even though it's a large concept, the work was laser-focused on keystone to allow feedback before a march larger investment of time was dedicated to the changing concept. i think this approach will help us all going forward.

larryrensing:
thanks for the feedback!

closing this out, we can figure out a node port structure for a values tree for a single service in a different pull request

Issue #200, opened 2017-02-14
feat: provide flexible operator overrides for all OpenStack configuration templates


alanmeadows:
This is a large feature that attempts to solve several critical items to make this project useful for real deployments:

1) Charts need a consistent and reliable way to offer operators the option to override configuration file settings.  For instance, if the operator knows they want to change the `debug` value in the `[DEFAULT]` section in `nova.conf`, what do they provide to `helm install --set` or in their own `--values` overrides, whether on the command line or in a large environment wide YAML.  Today, they would need to go inspect the chart and its templates to determine what this value is.  The proposal here would make this consistent, allowing the operator, once they understand the pattern to simply say `--set nova.conf.nova_conf.default.debug=true`

2) Operators need to be able to override more parameters then we supply in our configuration files today.  We provide a very small subset of the overall knobs [OpenStack has](http://docs.openstack.org/newton/config-reference/compute/config-options.html).  The proposal here would take the upstream configuration files for a service, and pull those in as our starting template.  We would then take every parameter in that file, and ensure that each line is something like:

```
[api_database]
mysql_sql_mode = `.Values.conf.nova_conf.api_database.mysql_sql_mode | default 'TRADITIONAL'
```
3) We should provide operators with the ability to arbitrary add lines at the end of any section by adding text to the following attribute:

```
.Values.conf.nova_conf.api_database.misc
```

4) We want to support operators providing their own pre-baked configuration files without having to fork our charts.  This means we would support the operator passing us their `nova.conf` verbatim.  Every template will need to begin with either using this (if its not nil) or using our standard out of the box template.

```
.Values.conf.nova_conf.override = """My raw config file"""
```

5) We want to provide operators the ability to arbitrary inject new files that our charts may not account for.  This means we need to refactor our configmaps across all charts to leverage `values.yaml` to understand what mounts to perform on what container.  This will also allow us to explicitly define read-only and ownership attriibutes as part of this effort.  With this in place, operators can inject additional mounts for `nova-api` for instance, if they need a `/etc/nova/my-custom-file.conf`

6) We will ensure all charts have all required files, including items like `policy.json`, `api-paste.json` and other ancilliary files so that any underlying image can be used that may or may not contain these dependencies.  Today, we have a needless reliance on them which also makes them immutable for the operator as they are baked into the image.

[...]

PR #262, opened 2017-03-09
Configuration Management Overrides for Keystone

alanmeadows:

This commit introduces several new paradigms:

* .Values driven mounts in service containers as well as jobs. These can be overriden at runtime by manipulating .Values.mounts.*; This would become a lot more convenient to append to if helm supported adding to an array via `--set` or in values.yaml.  Right now, it supports overriding it whole hog.

* Authenticated Endpoint Resolution methods in helm-toolkit to allow the database and rabbit type endpoints to be moved under the endpoints section.  Now endpoint lookup functions exist
for openstack endpoints as well as infrastructure elements such as the database, rabbitmq, and memcache.

* Importing of upstream "out of the box" policy.json and keystone-paste.ini files for better or worse. Helm will control these, not the image, just as it does with keystone.conf.

* A rendering of the keystone.conf oslo generated config using a new helm based Values generator. This provides "known" paths to all configuration elements in the form of `"conf.<file>.<section>.<oslo_namespace> <element>"` so that any of these can be targeted for override via `--set or values.yaml` files loaded at run time.  An example would be changing the keystone provider: `--set conf.keystone.token.keystone.provider=uuid`.  The file delineation provides operators a method to manipulate settings individually for charts that contain several components.  While it adds little value for keystone, for neutron for instance, you may want to enable debugging for the l3-agent, but not the neutron-server so a `.Values.conf.l3_agent` and `.Values.conf.neutron` prefix is required to support that.

* Support for appending arbitrary string data to the policy.json, keystone-paste.ini, or keystone.conf files via `--set conf.keystone.append="# some comment"`, or `--set conf.policy.append="\"identity:list_projects\": \"role:fake\""`

* Support for complete verbatim overrides of these three files via `--set conf.keystone.override="full file contents"`, or `--set conf.policy.override="full file contents"` for example

**What issue does this pull request address?**: Addresses #200 

@intlabs @v1k0d3n @DTadrzak @wilreichert @stannum-l 

Reviewers, please take a look at whether this approach (which can be done to every chart if this direction looks positive) fulfills large operator requirements.  Feedback is welcome.

@stannum-l you are tagged for your eagle eye toward Keystone correctness.


keystone/templates/etc/_policy.json.tpl
```diff
@@ -176,7 +184,7 @@
     "identity:list_projects_for_user": "",
     "identity:list_domains_for_user": "",
 
-    "identity:list_revoke_events": "rule:service_or_admin",
+    "identity:list_revoke_events": "",
```
stannum-l: do we want to relax the revocation_event_list policy here?

keystone/templates/etc/_keystone-paste.ini.tpl
```diff
@@ -74,17 +79,17 @@ use = egg:keystone#admin_service
 [pipeline:public_api]
 # The last item in this pipeline must be public_service or an equivalent
 # application. It cannot be a filter.
-pipeline = healthcheck cors sizelimit http_proxy_to_wsgi osprofiler url_normalize request_id build_auth_context token_auth json_body ec2_extension public_service
+pipeline = cors sizelimit http_proxy_to_wsgi osprofiler url_normalize request_id admin_token_auth build_auth_context token_auth json_body ec2_extension public_service
```
stannum-l: We should remove `admin_token_auth`.  Also, the `[filter:admin_token_auth]` at line 37 should be removed to avoid having `ADMIN_TOKEN` in the conf file.

keystone/templates/etc/_keystone-paste.ini.tpl
```diff
@@ -74,17 +79,17 @@ use = egg:keystone#admin_service
...
 
 [pipeline:admin_api]
 # The last item in this pipeline must be admin_service or an equivalent
 # application. It cannot be a filter.
-pipeline = healthcheck cors sizelimit http_proxy_to_wsgi osprofiler url_normalize request_id build_auth_context token_auth json_body ec2_extension s3_extension admin_service
+pipeline = cors sizelimit http_proxy_to_wsgi osprofiler url_normalize request_id admin_token_auth build_auth_context token_auth json_body ec2_extension s3_extension admin_service
```
stannum-l: ditto: remove `admin_token_auth`

keystone/templates/etc/_keystone-paste.ini.tpl
```diff
@@ -74,17 +79,17 @@ use = egg:keystone#admin_service
...
 
 [pipeline:api_v3]
 # The last item in this pipeline must be service_v3 or an equivalent
 # application. It cannot be a filter.
-pipeline = healthcheck cors sizelimit http_proxy_to_wsgi osprofiler url_normalize request_id build_auth_context token_auth json_body ec2_extension_v3 s3_extension service_v3
+pipeline = cors sizelimit http_proxy_to_wsgi osprofiler url_normalize request_id admin_token_auth build_auth_context token_auth json_body ec2_extension_v3 s3_extension service_v3
```
stannum-l: ditto: remove `admin_token_auth`.

keystone/templates/etc/_policy.json.tpl
stannum-l: FYI. a note is that starting in Pike, keystone will be removing this policy file in favor of moving policy management into code (similar to what nova is doing).

alanmeadows: It is not that I want to relax it, it is that I want to use the upstream Newton keystone files verbatim.

Unless it impedes our functionality, or opens a glaring security hole, I think we want to ingest these "as-is" to be providing a "familiar look and feel" to out-of-the-box.

keystone/templates/etc/_keystone-paste.ini.tpl
alanmeadows: @stannum-l To be sure, these are recommendations for our own deployment? (Which is good--we need this detail for that as well)

keystone/templates/etc/_policy.json.tpl
stannum-l: Ah - I just double checked the keystone repo, and I realized they changed that only in ocata.  I suspect it was an oversight in newton (question is do we want anyone besides project/admin to be able to list revocation events), but I agree it should be consistent with what's in the stable/newton branch.

keystone/templates/etc/_keystone-paste.ini.tpl
stannum-l: Yup.  Otherwise it is just really good practice to ask the community to migrate to use the bootstrap functionality rather than dealing with the `ADMIN_TOKEN.

intlabs: +1 to removal. `admin_token_auth`has no place in a new deployment, now bootstrapping can be performed without it.

keystone/templates/deployment.yaml
```diff
@@ -72,58 +74,8 @@ spec:
...
-            - name: mpmeventconf
-              mountPath: /etc/apache2/mods-available/mpm_event.conf
-              subPath: mpm_event.conf
-              readOnly: true
-            - name: startsh
-              mountPath: /tmp/start.sh
-              subPath: start.sh
-              readOnly: true
+          volumeMounts:            
+{{ toYaml $mounts.volumeMounts | indent 12 }}
```
intlabs: I really like the concept of this, nice work Alan. :)

However, I've a few concerns that (and there's a good chance I'm reading it wrong) this implies that we would be mounting the same volumes in each pod, regardless of functionality. If so - I think there may be value in defining volume mounts for each deployment/job class. Looking at the values.yaml I think we could also make this both a lot more compact and readable with a couple of minor tweaks, I'll try and knock up a gist/fork later today with a few of these thoughts fleshed out.

alanmeadows: This is not exactly the case.  Every pod does not receive the same mount.  That `$mounts` reference is declared early on:

```{{- $mounts := .Values.mounts.api }}```

There are other definitions for the jobs, and if this was a larger chart, for each deployment.

Now, that said, this does assume every container _within_ that pod would receive the same mounts, so we could deepen this tree and add a `mount.pod.container` definition to allow customizing it on a per container basis within the same pod.

I also looked at a way to make it more terse, but with the way it is now, as an operator you have maximum flexibility.  I can specify any mount option that exists because my YAML is passed unmolested into the manifest.  While we can come up with a DSL to shorten that, we need to be careful not to become like some products where the common subset is exposed, but advanced options are not exposed.

File: helm-toolkit/templates/_oslo.tpl [hunk 6]
```diff
@@ -0,0 +1,39 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+#-----------------------------------------------
+# oslo settings we will dynamically manufacture
+#-----------------------------------------------
+
+{{- define "helm-toolkit.oslo_values_setup" -}}
```
alanmeadows: My hope here is these (paths) will be consistent across all projects.  Corner cases may creep into here until we hassle oslo.

keystone/templates/bin/_init.sh.tpl
```diff
@@ -18,19 +18,19 @@ set -ex
 export HOME=/tmp
 
 ansible localhost -vvv \
-  -m mysql_db -a "login_host='{{ include "helm-toolkit.mariadb_host" . }}' \
-                  login_port='{{ .Values.database.port }}' \
-                  login_user='{{ .Values.database.root_user }}' \
-                  login_password='{{ .Values.database.root_password }}' \
-                  name='{{ .Values.database.keystone_database_name }}'"
+  -m mysql_db -a "login_host='{{ .Values.endpoints.oslo_db.hosts.internal | default .Values.endpoints.oslo_db.hosts.default }}' \
```
alanmeadows: This is only temporary until we ditch the kolla-ansible approach.  A separate PR would use the standard routines to lookup the endpoint and I would encourage that to support a traditional oslo.db URL.

keystone/templates/etc/_keystone-paste.ini.tpl
```diff
@@ -12,6 +12,14 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+{{- if empty .Values.conf.paste.override -}}
```
alanmeadows: Another alternative could be doing this logic in the config map definition (e.g. `configmap-etc.yaml` and `configmap-bin.yaml`) because overriding the contents, appending, or prepending is easily doable there.

The advantage of that is less logic within the template.

The templates did need to be wrapped in {{ defines }} which I dislike.  The purpose there is that without that, helm attempts to process them _prior_ to being included, and they lack the context necessary for them for the nested variables to be rendered.  It simply errors.  Wrapped in a define, it does not process until actually included.

I wish I did not have to introduce this as I like our files being "true to form" within etc/ and bin/ but I don't know what else to do other then bug @technosophos.

keystone/templates/etc/_keystone-paste.ini.tpl
```diff
@@ -109,3 +114,9 @@ use = egg:Paste#urlmap
 /v2.0 = admin_api
 /v3 = api_v3
 / = admin_version_api
+
+{{ if .append }}
```
alanmeadows: Again, an alternative is doing this sort of thing within `config-etc.yaml` and `config-bin.yaml` -- not within the template.

keystone/templates/etc/_policy.json.tpl
```diff
@@ -192,8 +200,14 @@
 
     "identity:create_domain_config": "rule:admin_required",
     "identity:get_domain_config": "rule:admin_required",
-    "identity:get_security_compliance_domain_config": "",
     "identity:update_domain_config": "rule:admin_required",
     "identity:delete_domain_config": "rule:admin_required",
     "identity:get_domain_config_default": "rule:admin_required"
+{{- if .append -}}
```
alanmeadows: Ugly, but it works.  It means the documentation needs to be clear that the operator must be injecting valid JSON that doesn't trail a comma.

keystone/values.yaml
```diff
@@ -108,6 +94,196 @@ resources:
...
+    - name: keystone-etc
+      mountPath: /etc/apache2/mods-available/mpm_event.conf
+      subPath: mpm_event.conf
+      readOnly: true
+    - name: keystone-bin
+      mountPath: /tmp/start.sh
+      subPath: start.sh
+      readOnly: true      
+
+conf:
```
alanmeadows: I dislike that this had to be done.  This is because nested invalid references in helm cannot be silenced.

{{ .Values.a.b.c }} where "c" doesn't exist is fine, as long as "b" is a valid map.  When its not, the template fails to render.  There is no terse way to gracefully check the existence of things and not b.get("c", None) concept which I have alluded to on https://github.com/kubernetes/helm/issues/2084

Thus, as a result, the entire oslo "skeleton" needs to be spelled out with empty leafs.

keystone/templates/deployment.yaml
intlabs: Cheers for clearing that up, will teach me for jumping to conclusions - I'm happy again :)

keystone/templates/bin/_init.sh.tpl
intlabs: @alanmeadows now the keystone endpoint work is merging, I'll get back on with the db management sans Ansible, and make a PR to get this in along with job standardization work across charts.

keystone/values.yaml
intlabs: Thinking out loud, would it be possible to create the required structure using the oslo-genconfig tool's output by inserting `set`s into it as appropriate?

keystone/templates/bin/_init.sh.tpl
v1k0d3n: :+1: @intlabs. i'll be on the lookout for it as well.

keystone/templates/etc/_keystone.conf.tpl
```diff
@@ -13,12 +13,122 @@
 # limitations under the License.
 
 {{- if empty .Values.conf.keystone.override -}}
+{{ include "keystone.conf.keystone_values_skeleton" .Values.conf.keystone | trunc 0 }}
 {{ include "keystone.conf.keystone" .Values.conf.keystone }}
 {{- else -}}
 {{ .Values.conf.keystone.override }}
 {{- end -}}
 
+{{- define "keystone.conf.keystone_values_skeleton" -}}
```
intlabs: Awesome :)

v1k0d3n: most definitely! :+1:

alanmeadows:
@stannum-l - I double checked, and this keystone.conf was generated against the newton branch.  Quite confident on that front.  I can only surmise they did not update the `stable/newton` sample with the `pre_settled` example.

alanmeadows:
I always support splitting to help complete a large effort but this is a
complete PR as opposed to a WIP,  and barring any critical feedback it is
ready to merge as is.

It leaves things in a completely functional manner, and the one liner
toward other charts is to required in the same pull request that exposes
mount options to the init container function which keystone needed.

It is the bare minimum of changes to introduce config management for
keystone alone.

Each chart will be a separate PR, so that will be the chunk approach
however once this model is approved I can make short order of all other
charts.

intlabs:
@alanmeadows @v1k0d3n This is a really impressive bit of work, and really moves us massively toward our goals. Like Brandon I've got some reservations about a PR of this size and Scope, but think that in this case the concepts introduced here make more sense as a cohesive whole. On that basis LGTM :)

@alanmeadows If you need assistance working through the remaining charts, please feel free to poke me.

larryrensing:
Awesome work @alanmeadows.  This looks really good, and provides a ton of flexibility.  To echo everyone else -- I usually dislike PRs of this size and scope, but I think this one is ready to go all at once.  The rest of the services should definitely be split up though.  LGTM

wilreichert:
Given the fairly intrusive nature of the changes would it be possible to schedule this to be merged at a predetermined time so we can plan the accordingly? That said I'd definitely like to see this get in ASAP.

alanmeadows:
@wilreichert Can you explain the need for that coordination, is there a CI/CD system pointed at master? Just trying to understand.  This merge will be swiftly followed up by PRs for every other chart to provide the same functionality.  We do want to get this in before the OpenStack github absorption.

flaper87:
I looked into this and the proposal looks good to me. I'll play a bit with it and provide more feedback but I wouldn't block this PR on it

Issue #258, opened 2017-03-08
Introduce 'development mode' consistently across all charts


wilkers-steve:
**Is this a bug report or feature request?** (choose one):  Feature request

**Kubernetes Version** (output of `kubectl version`):  Any

**Helm Client and Tiller Versions** (output of `helm version`): Any

**Development or Deployment Environment?**:  Development

**Release Tag or Master**:

**Expected Behavior**:  Ability to consistently set flags across all charts for developer mode to control the number of replicas where necessary and to circumvent the need for PVCs

**What Actually Happened**:

**How to Reproduce the Issue** (as minimally as possible):

**Any Additional Comments**:  

While some charts have the ability to set a development mode flag, others do not.  In order to maintain a positive developer experience for OpenStack-Helm going forward, we should consistently apply the ability to set a development mode flag across all future charts.

wilkers-steve:
I'll pick this up

v1k0d3n:
sounds good, it's yours. for item #111 that's already out there...are you going to take this on, or are we asking @intlabs to rebase one more time?

wilkers-steve:
@v1k0d3n, we can handle it however @intlabs wishes to -- if he'd like to walk it to completion, I'll exclude it from the work I'll submit against this issue.  If he'd like to avoid another rebase and not finish up #111, I'll add it to the list.

wilkers-steve:
While working on developer mode for the remainder of the service charts, I realized I wasn't entirely sure how to handle a developer mode setting for Cinder (admittedly, I'm not too familiar with Cinder) outside of spinning up a different replica count with a development flag.

Is there a way to handle developer mode here with hostPaths or emptyDirs that makes sense? I know Cinder needs a backend/backends enabled (whether it's Ceph, LVM, or something else), so would either of those two methods even work in this situation? I filed this issue so we can discuss the best path going forward.

Tagging @v1k0d3n @alanmeadows @intlabs for this one

alanmeadows:
@wilkers-steve I think development mode might be a good place to leverage the "Fake" backend:

https://github.com/openstack-dev/devstack/blob/master/lib/cinder_backends/fake

which simply logs the requests and allows you to create resources.

The other alternative, offering slightly more functionality (e.g. you can write bits and attach it to VMs if we eventually get nova working in development mode -- which is useful, albeit more effort to write, especially since someone pointed out this wasn't working at one point) would be using the LVM driver. 

[...]

PR #270, opened 2017-03-13
Add support for Cinder fake backend

wilkers-steve:
Use Cinder fake backend over Ceph for developer mode. This allows
developers to create volumes and get logging information for
testing purposes

**What is the purpose of this pull request?**:  Enable developers to use the Fake cinder backend driver for development purposes.  This eliminates reliance on Ceph for development

**What issue does this pull request address?**: Partially addresses #258

**Specific reviewers for pull request**: @alanmeadows @v1k0d3n @intlabs


cinder/values.yaml
```diff
@@ -22,6 +22,25 @@ replicas:
...
 
+backends:
+  enabled:
+  - rbd1
+  rbd1:
+    secret: null
+    user: "admin"
+    pool: "volumes"
+  fake:
+    enabled: false
```
v1k0d3n: i'd like to see what @alanmeadows recommends, but i would imagine a structure that could allow for multiple backends to be enabled simultaneously would be desired, like:

```
backends:
  fake:
    enabled: true
  rdb1:
    enabled: true
    secret: null
    user: "admin"
    pool: "volumes"
  lvm:
    enabled: true
...
```

docs/developer/minikube.md
```diff
@@ -189,12 +189,18 @@ To deploy Openstack-Helm in development mode, ensure you've created a minikube-a
...
 ```
 sudo mkdir -p /data/openstack-helm/mariadb
 ```
 
+
+### Change Cinder Backend
+
+Currently, Cinder uses Ceph for the RBD backend out of the box. To avoid Ceph in development mode, developers can use the Cinder 'Fake' backend by both adding the `fake` driver to the list of [enabled backends](https://github.com/att-comdev/openstack-helm/blob/master/cinder/values.yaml#L87) in Cinder's values.yaml file, and changing the `enabled` flag under `fake` to true. The Fake backend allows developers to create volumes and log events. In the future, the Fake backend will be replaced with the LVM backend in development mode to allow developers to create and attach volumes to instances.
+
+
```
v1k0d3n: i'd like to see a simpler one liner for development as an example (something cut and paste for developers), i.e.:

```
helm install --name cinder --set backend.fake.enabled=true --set backend.rdb1.enabled=false local/cinder --namespace=openstack
```

it just makes things so much easier for developers who are looking for quick cut/paste instructions.

cinder/values.yaml
wilkers-steve: I agree it looks a little unwieldy in its current state here. It can support multiple backends as is, but a developer would have to add another entry under the current 'enabled:' list. Currently there's a function in Helm-toolkit that handles dissecting these lists and entering them in the service configuration files, which relies on them being scoped how they are now. I can look into handling them in the manner you suggest because I agree, it does appear more readable.

docs/developer/minikube.md
wilkers-steve: Ok. Sounds good

v1k0d3n:
@wilkers-steve what's your plans with this? i think we need to still have the fake backend; it's just not something we're using as expected. is this a correct assumption? this way developers *could* use openstack-helm for development of backends, etc...like you brought up the other day.

if this work is scoped small, then we can open up a separate PR for LVM. i think this is how you were going to manage these; is that a correct assumption? if that is the case, let me know when you're ready to have this reviewed and I'll look it over again.

cinder/values.yaml
v1k0d3n: yup, ok...i get it. so a good thing to do would be to document the plans for how to use backends more generally (for anything). let me come up with a doc structure that i think would cover this well, and then we can tackle it more generally. sound good?

wilkers-steve:
After fiddling with the fake backend, I'm curious if it is necessary.  From what I've seen, it's used primarily for unit testing.  If a cinder developer wanted to develop a new backend, they'd do their cinder development then enable their new backend driver to test whether it works.  As far as I can tell, the fake backend wouldn't be necessary here

wilkers-steve:
The more important part for developing new backends here would be the docs you mentioned above and how to handle enabling multiple or different backends. Forgot to add that part 🙂

v1k0d3n:
well, if you want...we can close this out in favor of the LVM piece. your call on how you want to split out or drop the work.

wilkers-steve:
I'd be okay with dropping this PR in favor of LVM.  It'll give us more functionality and shouldn't be an insurmountable amount of work

v1k0d3n:
i'm ok with closing this in favor of switching to LVM. the main goal is to increase the capabilities of `development` mode, so if LVM opens that door...I'm all for it.

v1k0d3n:
just submit an issue/PR for LVM, and document how it can be used to improve development mode.

PR #271, opened 2017-03-14
bug: single rabbitmq instances don't need cluster readiness checks

wilreichert:
For single instance environments we're using the following overrides on the rabbitmq chart. It also avoids the extra etcd dependency.

```
auth:
  default_user: rabbitmq
dependencies:
  service:
enabled_plugins:
images:
  pull_policy: Always
replicas: 1
```

Unfortunately, the cluster readiness probe still runs every 10 seconds. The error is not fatal but is extremely noisy.

```
07:01:07 [readiness:4188] Starting readiness probe at 2017-03-13 21:57:39
07:01:07 [readiness:4188] Unexpected health-check output, giving the node the benefit of the doubt
07:01:07 [readiness:4188] Error: {{badmatch,{error,"Backend is not configured"}},
07:01:07 [readiness:4188]         [{autocluster,with_startup_lock,1,
07:01:07 [readiness:4188]                       [{file,"src/autocluster.erl"},{line,564}]},
07:01:07 [readiness:4188]          {autocluster,cluster_health_check_report,0,
07:01:07 [readiness:4188]                       [{file,"src/autocluster.erl"},{line,632}]},
07:01:07 [readiness:4188]          {autocluster,cluster_health_check,0,
07:01:07 [readiness:4188]                       [{file,"src/autocluster.erl"},{line,627}]},
07:01:07 [readiness:4188]          {erl_eval,do_apply,6,[{file,"erl_eval.erl"},{line,670}]},
07:01:07 [readiness:4188]          {rpc,'-handle_call_call/6-fun-0-',5,[{file,"rpc.erl"},{line,187}]}]}
07:01:07 [readiness:4188] Ready to return 0
07:01:07 [readiness:4685] Starting readiness probe at 2017-03-13 21:57:49
07:01:07 [readiness:4685] Unexpected health-check output, giving the node the benefit of the doubt
07:01:07 [readiness:4685] Error: {{badmatch,{error,"Backend is not configured"}},
07:01:07 [readiness:4685]         [{autocluster,with_startup_lock,1,
07:01:07 [readiness:4685]                       [{file,"src/autocluster.erl"},{line,564}]},
07:01:07 [readiness:4685]          {autocluster,cluster_health_check_report,0,
07:01:07 [readiness:4685]                       [{file,"src/autocluster.erl"},{line,632}]},
07:01:07 [readiness:4685]          {autocluster,cluster_health_check,0,
07:01:07 [readiness:4685]                       [{file,"src/autocluster.erl"},{line,627}]},
07:01:07 [readiness:4685]          {erl_eval,do_apply,6,[{file,"erl_eval.erl"},{line,670}]},
07:01:07 [readiness:4685]          {rpc,'-handle_call_call/6-fun-0-',5,[{file,"rpc.erl"},{line,187}]}]}
07:01:07 [readiness:4685] Ready to return 0
```

### DIFF
--- a/rabbitmq/templates/bin/_rabbitmq-readiness.sh.tpl
+++ b/rabbitmq/templates/bin/_rabbitmq-readiness.sh.tpl
@@ -33,7 +33,7 @@ main() {
         return 1
     fi
 
-    {{ if gt ".Values.replicas" "1" -}}
+    {{ if gt (.Values.replicas | int) 1 -}}
     if ! is-node-properly-clustered; then
         log-it "Node is inconsistent with the rest of the cluster"
         return 1
--- a/rabbitmq/templates/bin/_rabbitmq-readiness.sh.tpl
+++ b/rabbitmq/templates/bin/_rabbitmq-readiness.sh.tpl
@@ -32,10 +32,13 @@ main() {
         log-it "Node is unhealthy"
         return 1
     fi
+
+    {{ if gt ".Values.replicas" "1" -}}
     if ! is-node-properly-clustered; then
         log-it "Node is inconsistent with the rest of the cluster"
         return 1
     fi
+    {{- end }}
     return 0
 }
 


rabbitmq/templates/bin/_rabbitmq-readiness.sh.tpl
```diff
@@ -32,10 +32,13 @@ main() {
         log-it "Node is unhealthy"
         return 1
     fi
+
+    {{ if gt ".Values.replicas" "1" -}}
```
alanmeadows: I don't believe .Values.replicas should be quoted.

wilreichert: I've tried every other permutation quote & no quotes with those 2 values & that's the only one that doesn't give a `error calling gt: incompatible types for comparison` Is there a better way to handle this?

larryrensing: Some of the values conversions can be weird.  I'd personally try something like `{{ if gt (.Values.replicas | int) 1 }}`.  Casting the value to an int should do the trick

alanmeadows: That is a nice solution @larryrensing that future proofs against all sorts of nefarious possibilities, like users quoting integers in values.yaml.  

That said, to be sure, if `if gt .Values.replicas 1` isn't working (no quotes, anywhere) likely a bug needs to be filed with helm. I would legitimately expect an integer value in values.yaml (which is what `replicas: 3` should surface up as) to surface as a proper integer in the gotpl processing.

wilkers-steve: @alanmeadows, it looks like this is a result of Helm parsing every entry in values.yaml, including integers, as YAML strings and storing them all in a map of strings.  I'd suggest to just cast integers appropriately where needed for us in the interim, and we can file an issue with Helm (unless this is by design to make their lives easier).

I like @larryrensing's approach here.

gardlt: side-note: so talked to the helm guys `technosophos` and he said that this is an issue in how JSON handles int(sometimes they use float). and built-in conditions like `gt` `lt` only handle `int`. he recommended the same thing as @larryrensing.

wilkers-steve:
@wilreichert This looks good.  Thanks for making these changes :)

v1k0d3n:
looks good @wilreichert. thanks for the update!

PR #274, opened 2017-03-14
feat: Add NodePorts to heat service

larryrensing:
Add nodeports to heat service and ultimately decide on a structure for exposing nodeports for services
partial fix of #257 

this will serve as the nodeport values tree format for future work
@intlabs @v1k0d3n @alanmeadows

### DIFF
--- a/heat/templates/service-api.yaml
+++ b/heat/templates/service-api.yaml
@@ -19,5 +19,12 @@ metadata:
 spec:
   ports:
     - port: {{ .Values.network.api.port }}
+    {{ if .Values.network.api.node_port.enabled }}
+      nodePort: {{ .Values.network.api.node_port.port }}
+    {{ end }}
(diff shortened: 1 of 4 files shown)


intlabs:
In general, LGTM. Though I wonder should we document anywhere what the nodeports are to prevent collisions as we enable this for more services? An alternative would simply be to use the same nodeport as service port, meaning operators would either expand the kubeapi nodeport range or adjust as appropriate for their environment.

v1k0d3n:
@larryrensing can you create an issue for the continuation of `nodeport` conditionals, and work some someone on the team to complete the rest of this work; hopefully one of our newer team members in order get their feet wet? if we break them out per service it will be easier to chunk up, and we can have some other  community members start to participate in our repo as well. nice job establishing this needed pattern. :+1:

PR #284, opened 2017-03-15
bug: Add missing template for nova configmap-etc

### LINKED ISSUE #230: bug: Missing template for nova configmaps

wilkers-steve:
The ceph client cinder keyring template was missing from nova's
etc templates. This simply adds it in for the nova configmap to
consume

**What is the purpose of this pull request?**: Adds the missing ceph client cinder keyring template for nova

**What issue does this pull request address?**: Fixes #230 

**Specific reviewers for pull request**: @intlabs @larryrensing @v1k0d3n

### DIFF
--- a/nova/templates/etc/_ceph.client.cinder.keyring.yaml.tpl
+++ b/nova/templates/etc/_ceph.client.cinder.keyring.yaml.tpl
@@ -1,0 +1,20 @@
+# Copyright 2017 The Openstack-Helm Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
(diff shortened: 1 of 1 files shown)


intlabs:
LTGM @wilkers-steve. Though I think it would be great to get @alanmeadows input on this as well, I find it surprising we could have collectively missed such a seeming obvious element.

v1k0d3n:
@wilkers-steve were you able to test this? i'll have to pull in and test against a new ceph cluster. i agree, seems like a *major* oversight at first glance, and would be a bit worrying that something so obvious would be missed and not caught until now.

alanmeadows:
Looking closely at the references, it is not used by nova.conf as long as ceph is disabled, which it is by default.  This explains I think how this went unnoticed.  This change looks good to me.

PR #285, opened 2017-03-15
bug: Stnd control plane node issue #207

### LINKED ISSUE #207: Standardize control-plane node label value specification
**Is this a bug report or feature request?**: Mixed; depends on perspective

**Kubernetes Version** (output of `kubectl version`): N/A

**Helm Client and Tiller Versions** (output of `helm version`): N/A

**Development or Deployment Environment?**: N/A

**Release Tag or Master**: Master

**Expected Behavior**: As a developer and operator, it would be desirable to have standard control-plane node selector keys in the `values.yaml`. Currently, Nova uses a different set of keys for this functionality.

**What Actually Happened**: While editing charts, we cannot use the same standardized manifest for jobs as a result of some inconsistency with the way these values are specified.

**How to Reproduce the Issue** (as minimally as possible): N/A

**Any Additional Comments**: I suspect the values used in Nova are perfectly acceptable, and we should consider rolling this out to be the standard labeling for all charts. See the following for an example of how these currently differ:
 * Nova
   * [manifest template](https://github.com/att-comdev/openstack-helm/blob/master/nova/templates/job-db-init.yaml#L17)
   * [values](https://github.com/att-comdev/openstack-helm/blob/master/nova/values.yaml#L7-L8)
 * Keystone (and other charts)
   * [manifest template](https://github.com/att-comdev/openstack-helm/blob/master/keystone/templates/deployment.yaml#L29)
   * [values](https://github.com/att-comdev/openstack-helm/blob/master/keystone/values.yaml#L9-L10)

renmak:
Updating label key and value for openstack-control-plane. 

Was:
node_selector_key: openstack-control-plane
node_selector_value: enabled

Now/Update (within this PR)
control_node_selector_key: openstack-control-plane
control_node_selector_value: enabled

These changes applied to following
docs, glance, rabbitmq, postgresql, memcached, mariadb, maas, keystone, horizon, heat, etcd, cinder and neutron

**What is the purpose of this pull request?**: Standardizing openstack-control-plane label key and value

**What issue does this pull request address?**: Fixes #207 

**Specific reviewers for pull request**: @intlabs @v1k0d3n @alanmeadows

### DIFF
--- a/heat/templates/service-api.yaml
+++ b/heat/templates/service-api.yaml
@@ -19,5 +19,12 @@ metadata:
 spec:
   ports:
     - port: {{ .Values.network.api.port }}
+    {{ if .Values.network.api.node_port.enabled }}
+      nodePort: {{ .Values.network.api.node_port.port }}
+    {{ end }}
   selector:
     app: heat-api
+  {{ if .Values.network.api.node_port.enabled }}
+  type: NodePort
+  {{ end }}
+  
--- a/heat/templates/service-cfn.yaml
+++ b/heat/templates/service-cfn.yaml
@@ -19,5 +19,12 @@ metadata:
 spec:
   ports:
     - port: {{ .Values.network.cfn.port }}
+    {{ if .Values.network.cfn.node_port.enabled }}
+      nodePort: {{ .Values.network.cfn.node_port.port }}
+    {{ end }}
   selector:
     app: heat-cfn
+  {{ if .Values.network.cfn.node_port.enabled }}
+  type: NodePort
+  {{ end }}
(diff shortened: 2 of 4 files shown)


intlabs:
Nice work @renmak, and brings consistency for the top level key name. If possible it would be awesome if we could also look at the structure of the keys in Neutron (and potentially any others I've missed) so that we can get real uniformity across the charts? A pusdeo example to illustrate this:

```bash
helm install local/chart --set labels.control_node_selector_key=foo,labels.control_node_selector_value=bar
```

PR #287, opened 2017-03-16
docs: update minikube docs for non-writable systems

### LINKED ISSUE #95: feat/docs: Default values for development do not work on Atomic Linux
Many mount points including `/` are read-only on RPM-OSTREE Linux distros like Atomic, as a result, the default development values do not work: as there is no `/data` directory.

We should either update the documentation to tell users how to change this value, or update it to one that is acceptable for both minikube and Atomic Linux e.g: `/var/lib/localkube`.

* https://github.com/att-comdev/openstack-helm/blob/master/mariadb/values.yaml#L22

v1k0d3n:
**What is the purpose of this pull request?**: update documentation for atomic/coreos users who may want to develop using minikube.

**What issue does this pull request address?**: Fixes #95

**Notes for reviewers to consider**: not really

**Specific reviewers for pull request**: @intlabs @wilkers-steve @larryrensing


intlabs:
LGTM.

larryrensing:
LGTM, agree with @intlabs

v1k0d3n:
sure, the larger selection that talks about using the [`/data` dir](https://github.com/att-comdev/openstack-helm/blob/master/docs/developer/minikube.md#installation-and-testing) could be removed as part of this effort. it was originally taken from the pvc example in [minikube's docs](https://github.com/kubernetes/minikube#persistent-volumes), and we included some and would need to be changed in a few places like [HERE](https://github.com/att-comdev/openstack-helm/blob/6aaf950a49d239635200c71b61a6837d39e8b221/postgresql/values.yaml#L21) and [HERE](https://github.com/att-comdev/openstack-helm/blob/e7f5b6497ec4be23f8128a5f7745bf24c9f355b4/glance/values.yaml#L26) throughout the project.

but keeping this in proper scope, i'm not sure how many people are doing minikube development inside of Atomic/CoreOS. this flag only comes into play for [development mode](https://github.com/att-comdev/openstack-helm/blob/6aaf950a49d239635200c71b61a6837d39e8b221/postgresql/templates/deployment.yaml#L52-L56). this just seems like a real edge case to me, and  we are putting these advance users on notice.  i'll try to hit them later tonight if this is the ask.

v1k0d3n:
the more i think about this, i guess the more questions i have. it's easy to understand the reason for not wanting a `/data/` directory on systems like Atomic/CoreOS, however I just don't see these systems running Minikube over Kubernetes directly. Our "Development" docs clearly explain the intended use of Minikube. Because of this, `development` *could* (and potentially *should*) be renamed to `minikube`. Development/Minikube mode was driven towards the [PersistentVolumes](https://github.com/kubernetes/minikube#persistent-volumes) type `hostPath`, because this is something that should *just work* for developers out of the box. It was intentionally simplistic. We've recently had discussions on providing `development` alternatives, but I really think we need to nail down the most common use cases and iterate towards the desired end-state (without breaking `development` mode).

i'm going to close this PR and chose a path forward that we can more formally put on the roadmap. until we have something definitively agreed upon, we can direct people to Minikube `development` mode, or parts of the [installation/getting-started](https://github.com/v1k0d3n/openstack-helm/blob/master/docs/installation/getting-started.md) guide.

PR #288, opened 2017-03-16
Improve neutron chart with vxlan support.

korroot:
Added VLAN and VXLAN as possible tenant networking options.
Also, changed default tenant network type to VXLAN.
Creation of flat networking is still possible.

Improve neutron chart 
Add VLAN and VXLAN tenant network creation.
VXLAN tenant networks will be created as default. Other types need to be explicitly declared when creating network from CLI.

### DIFF
--- a/neutron/templates/etc/_ml2-conf.ini.tpl
+++ b/neutron/templates/etc/_ml2-conf.ini.tpl
@@ -17,7 +17,7 @@
 type_drivers = {{ include "helm-toolkit.joinListWithComma" .Values.ml2.type_drivers }}
 tenant_network_types = {{ .Values.ml2.tenant_network_types }}
 mechanism_drivers = {{ include "helm-toolkit.joinListWithComma" .Values.ml2.mechanism_drivers }}
-extension_drivers = {{ include "joinListWithComma" .Values.ml2.extension_drivers }}
+extension_drivers = {{ include "helm-toolkit.joinListWithComma" .Values.ml2.extension_drivers }}
 
 [ml2_type_flat]
 flat_networks = {{ include "helm-toolkit.joinListWithComma" .Values.ml2.ml2_type_flat.flat_networks }}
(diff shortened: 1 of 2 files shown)


neutron/templates/etc/_ml2-conf.ini.tpl
```diff
@@ -17,6 +17,7 @@
 type_drivers = {{ include "helm-toolkit.joinListWithComma" .Values.ml2.type_drivers }}
 tenant_network_types = {{ .Values.ml2.tenant_network_types }}
 mechanism_drivers = {{ include "helm-toolkit.joinListWithComma" .Values.ml2.mechanism_drivers }}
+extension_drivers = {{ include "joinListWithComma" .Values.ml2.extension_drivers }}
```
intlabs: this should be: `{{ include "helm-toolkit.joinListWithComma" .Values.ml2.extension_drivers }}`

v1k0d3n: +1

neutron/values.yaml
```diff
@@ -135,14 +135,18 @@ metadata:
   workers: 4
 
 ml2:
-  tenant_network_types: "flat"
+  tenant_network_types: "vxlan"
```
v1k0d3n: i'm confused, why are we changing the *defaults* in `values.yaml`?

korroot: In my use-case, I was deploying tenant network with vxlan as default. I can revert it to 'flat' if it is required.

intlabs:
@korroot I'd need to validate this later today, but LGTM. :)

neutron/values.yaml
```diff
@@ -137,12 +137,16 @@ metadata:
 ml2:
   tenant_network_types: "flat"
```
alanmeadows: I would of suspected this would still keep the flat network as the default (without testing):
```
[ml2]
...
tenant_network_types = {{ .Values.ml2.tenant_network_types }}
```

akorzeni: tenant_network_types will be used only when you do not specify the --provider:network_type. What exactly the default should be, it would be up to the user.

alanmeadows:
I'm good with getting this in -- this will be revisited when Configuration Overrides come to to the neutron chart in the next few days.  Getting this in now will mean that refactor will continue to ensure VXLAN is the default (which most people expect) and would be thoroughly tested as part of that upcoming refactor.

PR #273, opened 2017-03-14
feat: Update Nova Keystone Jobs to match common architecture

intlabs:
**What is the purpose of this pull request?**: To bring Nova Keystone Jobs inline with other services.

**What issue does this pull request address?**: N/A

**Notes for reviewers to consider**:
This PR should bring Nova Keystone Jobs inline with Cinder and Heat, and is part of the effort to standardize the common elements in charts prior to moving to OpenStack.

**Specific reviewers for pull request**:
@alanmeadows @v1k0d3n @larryrensing @wilkers-steve

Issue #201, opened 2017-02-14
feat: bootstrapping framework across all OpenStack charts


alanmeadows:
This feature requests that we support a consistent bootstrapping request framework across all charts.

Every OpenStack service has resources that can be created within the service once it is actually up.  Our charts deploy very vanilla versions of services, and to be sure, this is the correct path for us.  However, users wishing to deploy a functional cloud will need some basic items and we want to provide a mechanism to load the items at chart install.

A few examples:

Glance currently deploys with no images and Nova does not deploy with any usable flavors.  It should contain a `bootstrap` section in `values.yaml` containing the images to inject once the glance server is up (and the appropriate job to perform the action) and Nova should come up with usable flavors for deployment.

This is just an example:

```
bootstrap:
  images: 
    ubuntu: 
      name: Ubuntu Precise 64-bit
      source: http://cloud-images.ubuntu.com/precise/current/precise-server-cloudimg-amd64-disk1.img
      min_disk: 20
      image_type: qcow
```

Nova right now has no flavors, a `bootstrap` section in `values.yaml` could create the default out of the box flavors OpenStack users expect:

```
bootstrap:
  flavors:
    m1.tiny:
      name: m1.tiny
      id: auto
      ram: 512
      disk: 20
      vcpus: 1
```

intlabs:
In addition, there are other things we should be doing as part of the bootstrap process (off the top of my head):

- Creating and setting cinder volume types (required for some horizon workflows)
- Loading glance metadata schema (though we could include this in db migration)
- Create initial external network in Neutron
- Create appropriate initial User Roles in Keystone (and stop using admin for service accounts!)

wilreichert:
So as a matter of consistency is it worth applying a similar approach to the keystone chart? We've been discussing this & can see pros and cons either way. Adding bootstrap functionality means that now there's been an arbitrary determination of 'mandatory' and 'optional' configurations. One _needs_ a nova user for a functioning system but it will operate fine without a cirros image loaded. UNLESS you want to run tempest, in which case that optional image now becomes mandatory. I want to add a selenium helm test to the horizon chart which requires a couple users to exist in keystone, those are now mandatory for us but perhaps no one else. I guess the point is that whats mandatory or not will depend on the environment, so enforcing a strict definition might not make sense.

[...]

PR #290, opened 2017-03-16
feat: adding flavor bootstrap for nova

larryrensing:
This gives users an option to load nova with a basic flavor.

**What issue does this pull request address?**: partial fix for #201 

The setting of env vars in bootstrap.sh will change when #273 is merged

@intlabs @alanmeadows


nova/values.yaml
```diff
@@ -152,6 +164,17 @@ dependencies:
     - keystone-db-sync
     service:
     - mariadb
+  bootstrap:
+    jobs:
```
intlabs: Does the bootstrap job need to be dependent on any jobs itself? I think these can most likely all be safely removed, as it is only really dependent on the nova & keystone api's.

nova/values.yaml
```diff
@@ -152,6 +164,17 @@ dependencies:
...
+    - nova-db-init
+    - nova-post
+    - nova-db-sync
+    - keystone-db-init
+    - mariadb-seed
+    service:
+    - mariadb
```
intlabs: It should be safe to remove this dependency - as the bootstrap job will only be dependent on the api's themselves, and not the db backing them.

nova/values.yaml
```diff
@@ -47,6 +48,17 @@ upgrades:
     max_unavailable: 1
     max_surge: 3
 
+bootstrap:
+  enabled: true
+  flavors:
+    m1:
+      tiny:
```
intlabs: it would probably be better to have this as a single key, eg `m1_tiny` rather than a nest.

nova/templates/bin/_bootstrap.sh.tpl
```diff
@@ -0,0 +1,54 @@
...
+export OS_AUTH_STRATEGY=keystone
+export OS_REGION_NAME={{.Values.keystone.admin_region_name}}
+export OS_INSECURE=1
+EOF
+
+. /tmp/openrc
+
+wait_for_api
+
+nova flavor-show {{ .Values.bootstrap.flavors.m1.tiny.name }} || \
```
intlabs: would it be possible to expand this to an arbitrary number of flavors, specified in the values.yaml?

nova/templates/bin/_bootstrap.sh.tpl
```diff
@@ -0,0 +1,54 @@
...
+
+set -ex
+export HOME=/tmp
+
+
+function wait_for_api {
+	TIMEOUT=120
+	#larryrensing: sleep to wait until nova api is actually serving on
+	# the appropriate port.  not pretty, there has to be a cleaner approach to this
+	sleep 30
```
intlabs: Moving to direct service checking along the lines of: https://github.com/att-comdev/openstack-helm/pull/143 would remove this limitation of `kubernetes-entrypoint` to determine if a service is ready to receive requests. Until then is the sleep and the while statement needed or could we reduce it to one? If the job initially fails it presumably restarts wich may be a simpler, yet less pretty, way to deal with this.

nova/values.yaml
v1k0d3n: +1 would like see this as well. and this should really be something where company "acme" could add a list of flavors, no? even personally, i always have my 20 or so flavors after every single fresh installation.

nova/templates/bin/_bootstrap.sh.tpl
```diff
@@ -0,0 +1,54 @@
...
+# limitations under the License.
+
+set -ex
+export HOME=/tmp
+
+
+function wait_for_api {
+	TIMEOUT=120
+	#larryrensing: sleep to wait until nova api is actually serving on
+	# the appropriate port.  not pretty, there has to be a cleaner approach to this
```
v1k0d3n: were you planning on removing this statement, or just your github handle? i'd really like to remove the personal references (`notes` are fine). trying to catch them when we see them.

nova/values.yaml
larryrensing: agreed

nova/templates/job-bootstrap.yaml
```diff
@@ -0,0 +1,63 @@
...
+              mountPath: /tmp/bootstrap.sh
+              subPath: bootstrap.sh
+      volumes:
+        - name: novaconf
+          configMap:
+            name: nova-etc
+        - name: nova-bin
+          configMap:
+            name: nova-bin
+{{- end -}}
```
alanmeadows: See https://github.com/att-comdev/openstack-helm/pull/294 for comments on whether the job is conditional or the work is conditional.

nova/templates/bin/_bootstrap.sh.tpl
```diff
@@ -0,0 +1,47 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -ex
+export HOME=/tmp
+
+
+function wait_for_api {
```
alanmeadows: Why aren't we guaranteed this with the dependency checks? If we're not, we may want to investigate there.

nova/templates/bin/_bootstrap.sh.tpl
```diff
@@ -0,0 +1,47 @@
...
+	# note: sleep to wait until nova api is actually serving on
+	# the appropriate port.  not pretty, there has to be a cleaner approach to this
+	sleep 30
+	nova flavor-list 2>/dev/null
+}
+
+cat <<EOF>/tmp/openrc
+export OS_USERNAME={{.Values.keystone.admin_user}}
```
alanmeadows: Isn't there a helm-toolkit macro for this?

larryrensing: Yes - it was merged after this PR was submitted, a rebase is needed

nova/templates/bin/_bootstrap.sh.tpl
larryrensing: As @intlabs said on #294, the newest version of entrypoint may address the service dependency weirdness.

PR #294, opened 2017-03-20
feat: image bootstrap for glance

### LINKED ISSUE #201: feat: bootstrapping framework across all OpenStack charts

larryrensing:
Allow users to bootstrap a basic glance image

**What issue does this pull request address?**: Fixes #201 

@alanmeadows @intlabs @v1k0d3n @wilkers-steve

### DIFF
--- a/glance/templates/bin/_bootstrap.sh.tpl
+++ b/glance/templates/bin/_bootstrap.sh.tpl
@@ -1,0 +1,41 @@
+#!/bin/bash
+
+# Copyright 2017 The Openstack-Helm Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -ex
+export HOME=/tmp
+
+{{ if .Values.bootstrap.enabled }}
+
+cd /tmp/images
+
+{{ range .Values.bootstrap.images }}
+openstack image show {{ .name  | quote }} || \
+  { curl -O {{ .source_url }}{{ .image_file }}; \
+  openstack image create {{ .name | quote }} \
+  --min-disk {{ .min_disk }} \
+  --disk-format {{ .image_type }} \
+  --file {{ .image_file }} \
+  {{ if .properties -}} {{ range $key, $value := .properties }}--property {{$key}}={{$value}} {{ end }}{{ end -}} \
+  --container-format {{ .container_format | quote }} \
+  {{ if .private -}}
+  --private
+  {{- else -}}
+  --public
+  {{- end -}}; }
+{{ end }}
+
+{{ end }}
+exit 0
--- a/glance/templates/job-bootstrap.yaml
+++ b/glance/templates/job-bootstrap.yaml
@@ -1,0 +1,70 @@
+# Copyright 2017 The Openstack-Helm Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+{{- $envAll := . }}
+{{- $ksAdminSecret := $envAll.Values.keystone.admin_secret | default "glance-env-keystone-admin" }}
+{{- $dependencies := .Values.dependencies.bootstrap }}
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: glance-bootstrap
+spec:
+  template:
+    metadata:
+      annotations:
+        pod.beta.kubernetes.io/init-containers: '[
+{{ tuple $envAll $dependencies | include "helm-toolkit.kubernetes_entrypoint_init_container" | indent 10 }}
+        ]'
+    spec:
+      restartPolicy: OnFailure
+      nodeSelector:
+        {{ .Values.labels.node_selector_key }}: {{ .Values.labels.node_selector_value }}
+      containers:
+        - name: glance-bootstrap
+          image: {{ .Values.images.db_init }}
+          imagePullPolicy: {{ .Values.images.pull_policy }}
+          {{- if .Values.resources.enabled }}
+          resources:
+            limits:
+              cpu: {{ .Values.resources.jobs.bootstrap.limits.cpu | quote }}
+              memory: {{ .Values.resources.jobs.bootstrap.limits.memory | quote }}
+            requests:
+              cpu: {{ .Values.resources.jobs.bootstrap.requests.cpu | quote }}
+              memory: {{ .Values.resources.jobs.bootstrap.requests.memory | quote }}
+          {{- end }}
+          env:
+{{- with $env := dict "ksUserSecret" $ksAdminSecret }}
+{{- include "helm-toolkit.keystone_openrc_env_vars" $env | indent 12 }}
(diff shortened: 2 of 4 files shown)


glance/templates/bin/_bootstrap.sh.tpl
```diff
@@ -0,0 +1,40 @@
...
+}
+
+cd /tmp/images
+
+wait_for_api
+
+{{ range .Values.bootstrap.images }}
+openstack image show {{ .name  | quote }} || \
+  { curl -O {{ .source_url }}{{ .image_file }}; \
+  glance image-create --name {{ .name | quote }} --min-disk {{ .min_disk }} \
```
alanmeadows: Leveraging "openstack" client for testing and "glance" client to create.  We should stick with one pattern.

glance/templates/bin/_bootstrap.sh.tpl
```diff
@@ -0,0 +1,40 @@
...
+  --disk-format {{ .image_type }} \
+  --file {{ .image_file }} --container-format {{ .container_format }}; }
+{{ end }}
```
alanmeadows: What about other common need options such as public visibility, properties (e.g. NICs, disk types, etc.) and so on.

glance/templates/job-bootstrap.yaml
```diff
@@ -0,0 +1,72 @@
...
+      volumes:
+        - name: imagedir
+          emptyDir: {}
+        - name: bootstrapsh
+          configMap:
+            name: glance-bin
+        - name: glanceapiconf
+          configMap:
+            name: glance-etc
+{{- end -}}
```
alanmeadows: This could go two different ways: bootstrap job itself is conditional (dont create if not bootstrapping), or act of bootstrapping within job is conditional (e.g. if no bootstrapping, job just exits successfully doing nothing).  Just throwing this out there for discussion -- if the job is always there, you can potentially dependency check its completion.  I'm considering the use case of what if I want to wait until bootstrapping is complete, how would I do that with this arrangement.

glance/templates/bin/_bootstrap.sh.tpl
```diff
@@ -0,0 +1,40 @@
...
+
+	glance image-list 2>/dev/null
+}
+
+cd /tmp/images
+
+wait_for_api
+
+{{ range .Values.bootstrap.images }}
+openstack image show {{ .name  | quote }} || \
```
alanmeadows: Where is auth coming from?

glance/templates/bin/_bootstrap.sh.tpl
```diff
@@ -0,0 +1,40 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -ex
+export HOME=/tmp
+
+
+function wait_for_api {
```
alanmeadows: Shouldn't dependency checking guarantee this is ready? If not, seems like an underlying issue there.

glance/templates/bin/_bootstrap.sh.tpl
intlabs: OpenStack CLI would be preferable as the project specific CLI's are starting to fade away.

glance/templates/bin/_bootstrap.sh.tpl
intlabs: This should hopefully be resolved by: https://github.com/att-comdev/openstack-helm/pull/291, though will need to be tested.

glance/templates/bin/_bootstrap.sh.tpl
intlabs: @alanmeadows Looks like auth is being injected the same way it is for other keystone jobs: https://github.com/att-comdev/openstack-helm/pull/294/files/2a6a7f0a8ff265ade0a123ed935ccddd540b77b0#diff-3d0ea7bc2ea1e1bc18f6adea36a14a7aR48, though not sure if we want to use the cloud-admin, or service user here?

alanmeadows: Didn't catch that line. Not sure why https://github.com/att-comdev/openstack-helm/pull/290/files#diff-7d16c137cb5545d244cc1b5769d50f8bR30 diverges.

glance/templates/bin/_bootstrap.sh.tpl
v1k0d3n: +1 to @intlabs

glance/templates/job-bootstrap.yaml
larryrensing: This is a good point that I didn't consider- it would probably be best for dependencies to ensure that this job always finishes, even if it doesn't actually bootstrap anything.

glance/templates/bin/_bootstrap.sh.tpl
```diff
@@ -0,0 +1,41 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -ex
+export HOME=/tmp
+
+{{ if .Values.bootstrap.enabled }}
```
alanmeadows: Cool, so we'll get a successful job even if we aren't bootstrapping.

glance/templates/bin/_bootstrap.sh.tpl
```diff
@@ -0,0 +1,41 @@
...
+cd /tmp/images
+
+{{ range .Values.bootstrap.images }}
+openstack image show {{ .name  | quote }} || \
+  { curl -O {{ .source_url }}{{ .image_file }}; \
+  openstack image create {{ .name | quote }} \
+  --min-disk {{ .min_disk }} \
+  --disk-format {{ .image_type }} \
+  --file {{ .image_file }} \
+  {{ if .properties -}} {{ range $key, $value := .properties }}--property {{$key}}={{$value}} {{ end }}{{ end -}} \
```
alanmeadows: I'm glad we're account for properties as well.

glance/templates/bin/_bootstrap.sh.tpl
```diff
@@ -0,0 +1,41 @@
...
+set -ex
+export HOME=/tmp
+
+{{ if .Values.bootstrap.enabled }}
+
+cd /tmp/images
+
+{{ range .Values.bootstrap.images }}
+openstack image show {{ .name  | quote }} || \
+  { curl -O {{ .source_url }}{{ .image_file }}; \
```
alanmeadows: Likely we'll want to support additional methods of input in the future besides pulling it down but this is a good start.

larryrensing: I questioned this choice also -- as of glance api v2, passing in image URLs are no longer supported.  https://docs.openstack.org/developer/python-openstackclient/command-objects/image.html

We could mix v1 and v2 commands, but this functionality might be removed in the future.  At any rate, this is a good discussion item for the team as future bootstrapping service work may involve mixing API versions

intlabs: I think V1 will be gone with pike?

glance/templates/bin/_bootstrap.sh.tpl
wilkers-steve: I agree with @alanmeadows here -- we should look to include this ability as a next step once we move into OpenStack proper.  It'll be a good base for extending such functionality to the other services.

PR #300, opened 2017-03-22
LVM backend support for Cinder

### LINKED ISSUE #293: feat: Add support for Cinder LVM backend
**Is this a bug report or feature request?** (choose one): feature request

**Kubernetes Version** (output of `kubectl version`):

**Helm Client and Tiller Versions** (output of `helm version`):

**Development or Deployment Environment?**: Both

**Release Tag or Master**:

**Expected Behavior**:

**What Actually Happened**:

**How to Reproduce the Issue** (as minimally as possible):

**Any Additional Comments**:

After discussing desired functionality from Cinder in development mode in #258, we concluded that LVM support would be ideal.  Adding LVM support makes sense from a development point of view because it's functional and allows us to create volumes and attach them to VMs as opposed to leveraging Ceph, but LVM's also a backend that's used in production.  Adding support for LVM really tackles two issues, and I think it's a good step forward in the effort to remove development mode flags in our charts.  Documentation for using LVM in a local instance (laptops ideally) will accompany the work towards solving this issue.

wilkers-steve:
Initial commit for adding LVM support for Cinder. Adds necessary
configuration files for supporting configuration of the backend

**What is the purpose of this pull request?**: Add support for LVM backend for Cinder

**What issue does this pull request address?**: Fixes #293 

Currently the configmaps for the LVM backend aren't handled with conditionals.  This will change before the WIP tag gets removed.


v1k0d3n:
this is actually started as part of the development mode overrides (or improvement of development workflow), so it is an item on the 0.3.0 roadmap.

cinder/templates/etc/_cinder.conf.tpl
```diff
@@ -63,18 +69,30 @@ rabbit_password = {{ .Values.messaging.password }}
...
+#{{- else }}
+#rbd_secret_uuid = {{- include "secrets/ceph-client-key" . -}}
+#{{- end }}
+#rbd_secret_uuid = {{ .Values.backends.rbd1.secret }}
+#report_discard_supported = True
+
+[lvm1]
+iscsi_helper = tgtadm
+iscsi_protocol = iscsi
+lvm_conf_file = /etc/cinder/lvm.conf
```
gardlt: does this lvm.conf exist in the container, i dont see it in the template folder?

intlabs: It doesn't exist, I think we can safely remove the reference to it: https://github.com/openstack/cinder/blob/master/cinder/volume/drivers/lvm.py#L59-L64

cinder/templates/deployment-volume.yaml
```diff
@@ -43,6 +43,9 @@ spec:
         {{ .Values.labels.node_selector_key }}: {{ .Values.labels.node_selector_value }}
       containers:
         - name: cinder-volume
+          securityContext:
+             privileged: true
+             runAsUser: 0
```
intlabs: Now this is functional, we should be able to drop the privileges back down.

cinder/templates/etc/_cinder.conf.tpl
wilkers-steve: Yep.  This will be removed before the WIP tag comes off.

cinder/templates/etc/_cinder.conf.tpl
```diff
@@ -78,3 +85,18 @@ rbd_secret_uuid = {{- include "secrets/ceph-client-key" . -}}
 {{- end }}
 rbd_secret_uuid = {{ .Values.backends.rbd1.secret }}
 report_discard_supported = True
+{{ end }}
+
+{{ if .Values.lvm.enabled }}
```
alanmeadows: @intlabs @wilkers-steve  This is a comment for the future: 

Although some of this may change when cinder gets a config override pass, this section (and ceph above) to me says a helm-toolkit function to take a yaml declaration and turn it into ini format would be useful in places like this.  An equivalent of `{{ toYaml ...`` but geared towards oslo/ini.

This way you are not trying to provide overrides for every option possible, you give the end user the flexibility to target any known param, and add more.

Again, cinder config override stuff may account for some of this, but likely not all and we may still have a need to define backends much like Steve is doing here.  We won't know until we see what an expanded cinder.conf contains from oslo config gen.

In any event, what I am describing is:

values.yaml:
```
lvm:
  config:
    iscsi_helper: tgtadm
    lvm_conf_file: /etc/cinder/lvm.conf
    ...
```

And then here:

```
{{ if .Values.lvm.enabled }}
{{ tuple .Values.lvm.config | include "helm-toolkit.yaml_to_oslo" }}
{{ end }}
```

cinder/templates/etc/_rootwrap.conf.tpl
```diff
@@ -0,0 +1,27 @@
+# Configuration for cinder-rootwrap
```
alanmeadows: Good addition Steve. @v1k0d3n We likely need a feature to account for _rootwrap.conf.tpl consistency across the board.  This is the first (welcomed) introduction.  This is inline with me bringing in policy.json and paste everywhere.

cinder/values.yaml
```diff
@@ -153,6 +156,169 @@ dependencies:
     - keystone-api
     - cinder-api
 
+mounts:
```
alanmeadows: You did awesome laborious work here Steve.  There was a change of approach for mounts to make the .Values storage for them simply append as opposed to housing all mounts per https://github.com/att-comdev/openstack-helm/pull/319.  This means users cannot "undo" our  mounts but they can add it does make values.yaml a bit smaller and keeps the mounts we need closer to the actual manifest.  This is of course my fault for sending you down this path and then changing course after this PR.

We can either move these back into the manifests and copy https://github.com/att-comdev/openstack-helm/pull/319 in a separate PR or update them here.  Totally up to you.

cinder/templates/etc/_cinder.conf.tpl
wilkers-steve: @alanmeadows I like what you're describing.  I agree that'd be a useful feature to have on hand when required.

cinder/values.yaml
wilkers-steve: @alanmeadows Thanks for the feedback.  I can move these back to the manifests in this pull request.

wilkers-steve:
@alanmeadows @v1k0d3n @intlabs I've changed the mounts to function similarly to #319.  The only snowflake was cinder-volume, but we can address how to conditionally handle backend-specific mounts as we move forward

cinder/templates/deployment-volume.yaml
v1k0d3n: still running as root?

wilkers-steve: This will be removed with the next commit adding iSCSI to this PR.

v1k0d3n: well, to be fair...it's not the only one.i think privs and security is more of a project-wide security effort, and part of a later cleanup.

v1k0d3n:
@wilkers-steve is PR this in or out for openstack?

wilkers-steve:
@v1k0d3n I would say it could go either way.  I wouldn't let this block the -1 workflow on the patchset for moving into OpenStack.  If I get the iscsi bits added before the +2 happens for moving into infra, we can go ahead and merge it in.  If it doesn't, I'll add the work in once we move into OpenStack

wilkers-steve:
Lets get this in once we're in OpenStack.

v1k0d3n:

Issue #192, opened 2017-02-13
Developer environment fails to launch as a result of CNI issues within minikube.


intlabs:
**Is this a bug report or feature request?** (choose one): BUG REPORT

**Kubernetes Version** (output of `kubectl version`): v1.5.1 and v1.5.2

**Helm Client and Tiller Versions** (output of `helm version`): v2.1.3

**Development or Deployment Environment?**: Development

**Release Tag or Master**: All

**Expected Behavior**: Happy developers, rainbows, unicorns etc

**What Actually Happened**: Minikube blows up.

**How to Reproduce the Issue** (as minimally as possible):
Follow minikube developer documentation to the letter. (13/Feb/2017)

**Any Additional Comments**:

There appears to be some bugs in `minikube` when using CNI networking. In particular there looks to be a race issue with `ipset` that causes Calico to crash when loading a large number of pods: https://paste.fedoraproject.org/556865/94945148/raw/

As a workaround we can develop without CNI support, as we are not yet making use of network policy:

```
# Clone the project:
git clone https://github.com/att-comdev/openstack-helm.git && cd openstack-helm

# Get a list of the current tags:
git tag -l

# Checkout the tag you want to work with (if desired, or use master for development):
git checkout 0.1.0

# Start a local Helm Server:
helm serve &
helm repo add local http://localhost:8879/charts

# You may need to change these params for your environment. Look up use of --iso-url if needed:
minikube start \
        --kubernetes-version v1.5.2 \
        --disk-size 40g \
        --memory 16384 \
        --cpus 4 \
        --vm-driver kvm \
        --iso-url=https://storage.googleapis.com/minikube/iso/minikube-v1.0.6.iso

# Initialize Helm/Deploy Tiller:
helm init

# Package the Openstack-Helm Charts, and push them to your local Helm repository:
make

# Label the Minikube as an Openstack Control Plane node:
kubectl label nodes openstack-control-plane=enabled --all --namespace=openstack

# Deploy each chart:
helm install --name mariadb --set development.enabled=true local/mariadb --namespace=openstack
helm install --name=memcached local/memcached --namespace=openstack
helm install --name=rabbitmq local/rabbitmq --namespace=openstack
helm install --name=keystone local/keystone --namespace=openstack
helm install --name=cinder local/cinder --namespace=openstack
helm install --name=glance local/glance --namespace=openstack
helm install --name=heat local/heat --namespace=openstack
helm install --name=nova local/nova --namespace=openstack
[...]

Issue #268, opened 2017-03-13
feat: create custom minikube iso for openstack-helm development use


v1k0d3n:
**Kubernetes Version** (output of `kubectl version`): N/A
**Helm Client and Tiller Versions** (output of `helm version`): N/A
**Development or Deployment Environment?**: N/A
**Release Tag or Master**: N/A
**Expected Behavior**: N/A
**What Actually Happened**: N/A
**How to Reproduce the Issue** (as minimally as possible): N/A
**Any Additional Comments**:

A few times now, developers have experienced issues when upgrading Minikube. i'd like to get control of this, and build a known state iso image in order to control things that are needed for something like calico (which needs specific [kernel modules](https://github.com/kubernetes/minikube/issues/1007) included).

Use this as documentation for how to do this:
https://github.com/kubernetes/minikube/issues/1110

v1k0d3n:
i'll pick this up.

v1k0d3n:
@mcnanci this is really paused in favor of an issue that @intlabs said he would pick up: offering a new alternative development environment, that moves us away from minikube. i wouldn't say that this issue is blocked really, but there's no need to do work against it while the other effort is going on. there is no issue created for the new development alternative effort, but it is actively being explored: https://github.com/portdirect/dockerfiles/tree/kubeadm/aio-deploy/kubeadm-aio

v1k0d3n:
closing in favor of #dockerfiles/#76 and #304 .

PR #304, opened 2017-03-24
feat: Initial Vagrant/Kubeadm-aio development environment

intlabs:
**What is the purpose of this pull request?**: To provide a kubeadm based AIO development environment for OpenStack Helm

**What issue does this pull request address?**: Fixes https://github.com/att-comdev/openstack-helm/issues/192, and potentially replaces/supplements the approach proposed in https://github.com/att-comdev/openstack-helm/issues/268

@wilkers-steve @larryrensing @v1k0d3n

### DIFF
--- a/dev/README.md
+++ b/dev/README.md
@@ -1,0 +1,43 @@
+# Development Environment Setup
+
+## Requirements
+
+  * Hardware
+    * 16GB RAM
+    * 32GB HDD Space
+  * Software
+    * Vagrant
+    * VirtualBox
+    * Kubectl
+    * Helm
+    * Git
+
+## Deploy
+
+  * Make sure you are in the directory containing the Vagrantfile before running the following commands.
+
+### Create VM
+
+``` bash
+vagrant up --provider virtualbox
+```
+
+### Deploy NFS Provisioner for development PVCs
+
+``` bash
+vagrant ssh --command "sudo docker exec kubeadm-aio kubectl create -R -f /opt/nfs-provisioner/"
+```
+
+### Setup Clients and deploy Helm's tiller
+
+``` bash
+./setup-dev-host.sh
+```
+
+### Label VM node(s) for OpenStack-Helm Deployment
+
+``` bash
+kubectl label nodes openstack-control-plane=enabled --all --namespace=openstack
+kubectl label nodes openvswitch=enabled --all --namespace=openstack
+kubectl label nodes openstack-compute-node=enabled --all --namespace=openstack
+```
--- a/dev/README.md
+++ b/dev/README.md
@@ -6,8 +6,8 @@
     * 16GB RAM
     * 32GB HDD Space
   * Software
-    * Vagrant
-    * VirtualBox
+    * Vagrant >= 1.8.0
+    * VirtualBox >= 5.1.0
     * Kubectl
     * Helm
     * Git
--- a/dev/Vagrantfile
+++ b/dev/Vagrantfile
@@ -21,7 +21,7 @@ Vagrant.configure("2") do |config|
 
   # Every Vagrant development environment requires a box. You can search for
   # boxes at https://atlas.hashicorp.com/search.
-  config.vm.box = "ubuntu/xenial64"
+  config.vm.box = $vm_image
 
   # Disable automatic box update checking. If you disable this, then
   # boxes will only be checked for updates when the user runs
@@ -42,7 +42,7 @@ Vagrant.configure("2") do |config|
   # backing providers for Vagrant. These expose provider-specific options.
   config.vm.provider "virtualbox" do |vb|
     # Display the VirtualBox GUI when booting the machine
-    vb.gui = true
+    vb.gui = false
 
     # Customize the amount of memory on the VM:
     vb.memory = $ram
@@ -51,10 +51,10 @@ Vagrant.configure("2") do |config|
     vb.cpus = $vcpu_cores
 
     # Set the size of the VM's root disk:
-    unless File.exist?('.vagrant/machines/default/virtualbox/openstack-helm-storage.vdi')
-      vb.customize ['createhd', '--filename', '.vagrant/machines/default/virtualbox/openstack-helm-storage', '--size', $docker_disk]
+    unless File.exist?('.vagrant/machines/default/openstack-helm-storage.vdi')
+      vb.customize ['createhd', '--filename', '.vagrant/machines/default/openstack-helm-storage', '--size', $docker_disk]
     end
-    vb.customize ['storageattach', :id, '--storagectl', 'SCSI', '--port', 2, '--device', 0, '--type', 'hdd', '--medium', '.vagrant/machines/default/virtualbox/openstack-helm-storage.vdi']
+    vb.customize ['storageattach', :id, '--storagectl', 'SCSI', '--port', 2, '--device', 0, '--type', 'hdd', '--medium', '.vagrant/machines/default/openstack-helm-storage.vdi']
 
   end
 
--- a/dev/Vagrantfile
+++ b/dev/Vagrantfile
@@ -1,0 +1,91 @@
+# -*- mode: ruby -*-
+# vi: set ft=ruby :
+
+# All Vagrant configuration is done below. The "2" in Vagrant.configure
+# configures the configuration version (we support older styles for
+# backwards compatibility). Please don't change it unless you know what
+# you're doing.
+Vagrant.configure("2") do |config|
+  # The most common configuration options are documented and commented below.
+  # For a complete reference, please see the online documentation at
+  # https://docs.vagrantup.com.
+
+  # Every Vagrant development environment requires a box. You can search for
+  # boxes at https://atlas.hashicorp.com/search.
+  config.vm.box = "ubuntu/xenial64"
+
+  # Disable automatic box update checking. If you disable this, then
+  # boxes will only be checked for updates when the user runs
+  # `vagrant box outdated`. This is not recommended.
+  # config.vm.box_check_update = false
+
+  # Create a forwarded port mapping which allows access to a specific port
+  # within the machine from a port on the host machine. In the example below,
+  # accessing "localhost:8080" will access port 80 on the guest machine.
+  config.vm.network "forwarded_port", guest: 6443, host: 8443
+
+  # Create a private network, which allows host-only access to the machine
+  # using a specific IP.
+  # config.vm.network "private_network", ip: "192.168.33.10"
+
+  # Create a public network, which generally matched to bridged network.
+  # Bridged networks make the machine appear as another physical device on
+  # your network.
+  # config.vm.network "public_network"
+
+  # Share an additional folder to the guest VM. The first argument is
+  # the path on the host to the actual folder. The second argument is
+  # the path on the guest to mount the folder. And the optional third
+  # argument is a set of non-required options.
+  # config.vm.synced_folder "../data", "/vagrant_data"
+
+  # Provider-specific configuration so you can fine-tune various
+  # backing providers for Vagrant. These expose provider-specific options.
+  # Example for VirtualBox:
+  #
+  config.vm.provider "virtualbox" do |vb|
+    # Display the VirtualBox GUI when booting the machine
+    vb.gui = true
+
+    # Customize the amount of memory on the VM:
+    vb.memory = "8192"
+  end
+  #
+  # View the documentation for the provider you are using for more
+  # information on available options.
+
+  # Define a Vagrant Push strategy for pushing to Atlas. Other push strategies
+  # such as FTP and Heroku are also available. See the documentation at
+  # https://docs.vagrantup.com/v2/push/atlas.html for more information.
+  # config.push.define "atlas" do |push|
+  #   push.app = "YOUR_ATLAS_USERNAME/YOUR_APPLICATION_NAME"
+  # end
+
+  # Enable provisioning with a shell script. Additional provisioners such as
+  # Puppet, Chef, Ansible, Salt, and Docker are also available. Please see the
+  # documentation for more information about their specific syntax and use.
+  config.vm.provision "shell", inline: <<-SHELL
+     apt-get update
+     apt-get install -y \
+        docker.io \
+        nfs-common
+      mkdir -p /var/lib/kublet
+      mount --bind /var/lib/kublet /var/lib/kublet
+      mount --make-shared /var/lib/kublet
+      docker run \
+        -dt \
+        --name=kubeadm-aio \
+        --net=host \
+        --security-opt=seccomp:unconfined \
+        --cap-add=SYS_ADMIN \
+        --tmpfs=/run \
+        --tmpfs=/run/lock \
+        --volume=/etc/machine-id:/etc/machine-id:ro \
+        --volume=/home:/home:rw \
+        --volume=/etc/kubernetes:/etc/kubernetes:rw \
+        --volume=/sys/fs/cgroup:/sys/fs/cgroup:ro \
+        --volume=/var/run/docker.sock:/run/docker.sock \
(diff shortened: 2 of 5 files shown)


dev/Vagrantfile
```diff
@@ -0,0 +1,91 @@
...
+        --cap-add=SYS_ADMIN \
+        --tmpfs=/run \
+        --tmpfs=/run/lock \
+        --volume=/etc/machine-id:/etc/machine-id:ro \
+        --volume=/home:/home:rw \
+        --volume=/etc/kubernetes:/etc/kubernetes:rw \
+        --volume=/sys/fs/cgroup:/sys/fs/cgroup:ro \
+        --volume=/var/run/docker.sock:/run/docker.sock \
+        --env KUBELET_CONTAINER=docker.io/port/kubeadm-aio:latest \
+        docker.io/port/kubeadm-aio:latest
```
wilkers-steve: ￼ +        docker.io/port/kubeadm-aio:latest
￼ 
￼￼  ￼Is the Dockerfile at a point it can be tagged with something other than latest?  I vaguely recall you mentioning you needed to address some bugs with it, but a tagged version would give us a consistent base to use for now

v1k0d3n: agreed. would like to see this move towards something that's supported out of our quay repo, and eventually openstack (just another reason for openstack to have a registry). this way, the team could manage the container creation/versioning.

dev/Vagrantfile
```diff
@@ -0,0 +1,91 @@
...
+# backwards compatibility). Please don't change it unless you know what
+# you're doing.
+Vagrant.configure("2") do |config|
+  # The most common configuration options are documented and commented below.
+  # For a complete reference, please see the online documentation at
+  # https://docs.vagrantup.com.
+
+  # Every Vagrant development environment requires a box. You can search for
+  # boxes at https://atlas.hashicorp.com/search.
+  config.vm.box = "ubuntu/xenial64"
```
v1k0d3n: we may want to configure an `overrides.rb` file so we can use variables for things like memory and `kubeadm-aio:$version`. we could literally take a *large* chunk from halcyon to make this useful for vbox and libvirt.

dev/Vagrantfile
```diff
@@ -0,0 +1,91 @@
...
+  # such as FTP and Heroku are also available. See the documentation at
+  # https://docs.vagrantup.com/v2/push/atlas.html for more information.
+  # config.push.define "atlas" do |push|
+  #   push.app = "YOUR_ATLAS_USERNAME/YOUR_APPLICATION_NAME"
+  # end
+
+  # Enable provisioning with a shell script. Additional provisioners such as
+  # Puppet, Chef, Ansible, Salt, and Docker are also available. Please see the
+  # documentation for more information about their specific syntax and use.
+  config.vm.provision "shell", inline: <<-SHELL
```
v1k0d3n: shell provisioner is nice as a way to free ourselves of an ansible requirement. i like it. if we get requests for RH, then ansible could start to creep back in.

intlabs: I don't think this is likely, this method already supports redhat hosts.

dev/Vagrantfile
intlabs: One I've got Virtualbox support rock solid I'll bring this in.

intlabs: @v1k0d3n I've now put a `config.rb` in in the same manner as halcyon - Once we've got feedback that this works ok then we can add support for libvirt.

intlabs:
Hey this is ready for your collective expert gaze :) The docs are severely lacking atm, but would appreciate some guidance on how people would like me to tackle this.

dev/Vagrantfile
```diff
@@ -0,0 +1,93 @@
...
+        --tmpfs=/run \
+        --tmpfs=/run/lock \
+        --volume=/etc/machine-id:/etc/machine-id:ro \
+        --volume=/home:/home:rw \
+        --volume=/etc/kubernetes:/etc/kubernetes:rw \
+        --volume=/sys/fs/cgroup:/sys/fs/cgroup:ro \
+        --volume=/var/run/docker.sock:/run/docker.sock \
+        --env KUBE_BIND_DEV=enp0s8 \
+        --env KUBELET_CONTAINER=docker.io/port/kubeadm-aio:latest \
+        docker.io/port/kubeadm-aio:latest
```
intlabs: I'll make a pr for the image, and then update as soon as we have a tag/ref to point too.

intlabs:
This PR depends on the image here being merged/published: https://github.com/att-comdev/dockerfiles/pull/82

dev/README.md
```diff
@@ -0,0 +1,43 @@
+# Development Environment Setup
+
```
RichWellum: Missing a description along the lines of https://github.com/portdirect/dockerfiles/tree/kubeadm/aio-deploy/kubeadm-aio - 'This container builds a small AIO Kubeadm based Kubernetes deployment for Demo and Development use. For convenience helm is included in the image and a script to bring up a development OpenStack-Helm environment.'

dev/README.md
```diff
@@ -0,0 +1,43 @@
...
+  * Hardware
+    * 16GB RAM
+    * 32GB HDD Space
+  * Software
+    * Vagrant
+    * VirtualBox
+    * Kubectl
+    * Helm
+    * Git
+
```
RichWellum: I've been running the AIO manually in a VM, as I don't have unfettered access to a physical host. Using vagrant seems to make this heavy because you'd have a lot of nested virtualization going on: host->vm->vagrant-vm->docker. So maybe under Requirements describe the use cases? 'Laptop, bare-metal server etc'?.

dev/Vagrantfile
```diff
@@ -0,0 +1,93 @@
+# -*- mode: ruby -*-
```
RichWellum: I'm trying to understand why vagrant. I'm a big vagrant fan, it's great for windows users, and also for bringing up a VM pre-configured for certain purposes. In your case though, you are using vagrant mainly to pull a VM (that part is fine) then run docker automatically. I know you're going to add centos-libvirt later, but what about VM->ubuntu/centos->docker use-case. I feel that you're basing the whole AIO on vagrant, whereas it should be an offering. The programmatic part of the Vagrantfile - well https://github.com/portdirect/dockerfiles/tree/kubeadm/aio-deploy/kubeadm-aio covers this and it's so simple - I'm not sure vagrant is something I want to run. 

Maybe I'm reading this all wrong - apologizing in advance if that's the case!

dev/Vagrantfile
```diff
@@ -0,0 +1,93 @@
...
+
+  # Enable provisioning with a shell script.
+  config.vm.provision "shell", inline: <<-SHELL
+     # Setup docker storage
+     mkfs.xfs /dev/disk/by-path/pci-0000\:00\:14.0-scsi-0\:0\:2\:0  -f -L docker-srg
+     mkdir -p /var/lib/docker
+     echo "LABEL=docker-srg /var/lib/docker xfs defaults 0 0" >> /etc/fstab
+     mount -a
+
+     apt-get update
```
RichWellum: Just pointing out this is not centos/yum compatible. I know know you this, but maybe logic to do 'lsb_release -a' and if it's not ubuntu quit with a friendly message.

dev/Vagrantfile
```diff
@@ -0,0 +1,93 @@
...
+        --security-opt=seccomp:unconfined \
+        --cap-add=SYS_ADMIN \
+        --tmpfs=/run \
+        --tmpfs=/run/lock \
+        --volume=/etc/machine-id:/etc/machine-id:ro \
+        --volume=/home:/home:rw \
+        --volume=/etc/kubernetes:/etc/kubernetes:rw \
+        --volume=/sys/fs/cgroup:/sys/fs/cgroup:ro \
+        --volume=/var/run/docker.sock:/run/docker.sock \
+        --env KUBE_BIND_DEV=enp0s8 \
```
RichWellum: Along the lines of our IRC conversation - this will have to be unique to the users environment? Or is vagrant taking care of this?

dev/setup-dev-host.sh
```diff
@@ -0,0 +1,17 @@
...
+fi
+echo "Getting kubeconfig from kube1"
+vagrant ssh default -c "sudo cat /etc/kubernetes/admin.conf" > ${HOME}/.kube/config
+
+# Setting up helm client if present
+if which helm 2>/dev/null; then
+  helm init
+fi
+
+echo "clients should now be ready to access the Kubernetes cluster"
```
RichWellum: Maybe add echo "Use: 'sudo docker exec -it kubeadm-aio bash'" to access AIO Cluster.

dev/Vagrantfile
intlabs: @RichWellum This is the vagrant implementation of that container, so that we can get away from using minikube. It's not intended to replace that container at all - I'm running it directly on my laptop for example, and don't bother with a VM at all (Be warned, there is a *HIGH* probability of terrible things happening if you don't know exactly what's going on with this stuff).

dev/Vagrantfile
intlabs: This will come at a later point, once we have ensured that the UX is what we expect, it's why I did not move the image out to `config.rb` just yet :)

dev/Vagrantfile
intlabs: This will be the same for every deployment using Vagrant/Virtualbox and Ubuntu's predictable device naming.

dev/setup-dev-host.sh
intlabs: They should now just be able to use clients on the host, there is no need to enter the VM at all :)

dev/README.md
intlabs: Vagrant would not be running on a VM, but on the developers host. I'll update to note that this has been tested on MacOS, Ubuntu and Fedora workstations.

RichWellum: Ah the comment below: "### Create VM" - was confusing me a little as it's been a while since I played with vagrant.

dev/Vagrantfile
RichWellum: And that doesn't make you want to run it in a VM? Then it's isolated and safe.

dev/Vagrantfile
RichWellum: Thanks.

dev/Vagrantfile
RichWellum: Thanks for confirming. But of course the dependency is still on ubuntu being the host OS. Which to me is slightly against the vagrant model. Whereas if you were running it in a vagrant ubuntu vm.... ;)

dev/setup-dev-host.sh
RichWellum: Got it - thanks.

dev/Vagrantfile
v1k0d3n: you can use `latest`, but for more granularity, the following is now available: `quay.io//attcomdev/kubeadm-aio:v1.0.0`

dev/README.md
v1k0d3n: @intlabs just to be clear, vagrant is available for windows too, but the obvious gap would be provisioning would break (just have to let users know to do things manually). that's a future documentation item, which hopefully someone in the community can pick up.

dev/Vagrantfile
v1k0d3n: @RichWellum these are considerations of course. i think @intlabs mentioned that the `config.rb` would follow a similar path to our halcyon repo which seems to handle centos/ubuntu just fine. these things are all configuration knobs that we'll need to keep on the radar. my expectation is that you and the rest of the community (soon in openstack) will keep the projects honest. :)

dev/setup-dev-host.sh
v1k0d3n: @intlabs this would rewrite any ~./kube/config the developer may have for other environments. we may want to just add the contents of the vagrant host `admin.conf` as a kubeadm-aio context (which is set to enabled). thoughts on doing this with a little less disruption?

dev/README.md
intlabs: @v1k0d3n of course :) though I'm not able to test on windows atm, and this also relies on bash being present on the host, so not sure how that works with the compatibility layer. If someone was able to pick up the gaps here I'd really apprecate it,

dev/setup-dev-host.sh
intlabs: This is the exact same approach we took with halcyon - where the ${HOME}/.kube/config is backed up prior to being substituted with the one for the dev environment.

v1k0d3n: i can live with that. i think we should look at plucking items out of the `admin.conf`, and setting up as a separate context in the very near future. doesn't have to be in this PR, as i'd like to merge this in now (all my questions have been answered), but I would like to track this as an improvement to the development environment we're offering our community.

RichWellum: :+1: LGTM thanks for the explanation(s).

dev/Vagrantfile
```diff
@@ -0,0 +1,93 @@
...
+  # the path on the host to the actual folder. The second argument is
+  # the path on the guest to mount the folder. And the optional third
+  # argument is a set of non-required options.
+  config.vm.synced_folder "../", "/opt/openstack-helm"
+
+  # Provider-specific configuration so you can fine-tune various
+  # backing providers for Vagrant. These expose provider-specific options.
+  config.vm.provider "virtualbox" do |vb|
+    # Display the VirtualBox GUI when booting the machine
+    vb.gui = true
```
gardlt: I think it would be more preferable to run this in headless mode.

dev/config.rb
```diff
@@ -0,0 +1,4 @@
+# VM Specs
```
gardlt: is there are a reason to put these vars in this file instead of the Vagrantfile? If there is, we should put a note on the README.md so they can mod it later.

v1k0d3n: @gardlt yeah, the `config.rb` is a one-stop place for users to override values, so the actual Vagrantfile can remain essentially untouched. it's the same way that Halcyon is delivered today.

dev/Vagrantfile
v1k0d3n: +1 unless there is a strong reason for doing otherwise.

dev/README.md
```diff
@@ -0,0 +1,43 @@
+# Development Environment Setup
+
+## Requirements
+
+  * Hardware
+    * 16GB RAM
+    * 32GB HDD Space
+  * Software
+    * Vagrant
```
gardlt: For Vagrant and VirtualBox, could you put some suggested version? I was running into some issues in an old ver of VMbox on OSX. when updated it worked fine.

intlabs: Done

dev/Vagrantfile
```diff
@@ -0,0 +1,93 @@
...
+      vb.customize ['createhd', '--filename', '.vagrant/machines/default/openstack-helm-storage', '--size', $docker_disk]
+    end
+    vb.customize ['storageattach', :id, '--storagectl', 'SCSI', '--port', 2, '--device', 0, '--type', 'hdd', '--medium', '.vagrant/machines/default/openstack-helm-storage.vdi']
+
+  end
+
+  # Enable provisioning with a shell script.
+  config.vm.provision "shell", inline: <<-SHELL
+     # Setup docker storage
+     mkfs.xfs /dev/disk/by-path/pci-0000\:00\:14.0-scsi-0\:0\:2\:0  -f -L docker-srg
```
alanmeadows: Presumably apt-get install xfs-utils (and eventually its yum counterpart) is cared for somewhere or a dependency of vagrant?

intlabs: This command runs inside the VM, and `xfs-utils` is included by default in the vagrant images.

v1k0d3n:
@intlabs nice NFS additions. LGTM.

gardlt:
This LGTM. I been using this vagrantfile, i have been able to deploy keystone.

PR #318, opened 2017-03-31
Feat: bootstrap services cinder

jeffaugustine:
Initialize Cinder with a volume type

partial fix for #201  

@larryrensing @intlabs

### DIFF
--- a/cinder/templates/_bootstrap.sh.tpl
+++ b/cinder/templates/_bootstrap.sh.tpl
@@ -1,3 +1,5 @@
+#!/bin/bash
+
 # Copyright 2017 The Openstack-Helm Authors.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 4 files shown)


larryrensing:
After thinking about it for a while and bouncing some ideas off others, is this something we really want for a service like cinder?  With the work @wilkers-steve is doing [here](https://github.com/att-comdev/openstack-helm/pull/300), this bootstrapping job would fail if LVM is enabled with ceph disabled.  These bootstrapping jobs should be idempotent regardless of what configurations have been made under the hood.  I think cinder is too much of a snowflake to be able to bootstrap with confidence.  Thoughts @alanmeadows @intlabs @wilkers-steve @v1k0d3n ?

intlabs:
With this implementation, I'm inclined to agree with you @larryrensing. It would, however, be possible to get the enabled backends from the cinder.config and then bootstrap the appropriate volume types from there.

wilkers-steve:
Thanks for working this @jeffaugustine.  I agree with @larryrensing and @intlabs that we should reevaluate how we're bootstrapping volume types for cinder.  One of the necessary additions for bootstrapping will require polling whether a resource has been successfully created and has a status of `ready`.  I'd like to revisit how Cinder can handle bootstrapping in an idempotent way that will account for all enabled/desired backends instead of with a single top level `enabled` flag.  If @alanmeadows @larryrensing and @intlabs agree, I'd like to see this handled after the move to OpenStack once we have that polling functionality figured out and have a more clear direction on multiple backends.

v1k0d3n:
i'm in favor of waiting until openstack @jeffaugustine. this will give you some breathing room (you don't have to feel rushed), and allow the team to really discuss it further. +1 to @wilkers-steve

jeffaugustine:
Alright. I'll close for now with intent to reintroduce once everything is moved over, using a more dynamic approach to choosing what volume type(s) is created.

PR #319, opened 2017-03-31
Nova Configuration Override

alanmeadows:

    Initial Commit of Nova Configuration Overrides
    
    This commit is an evolution of the initial keystone configuration
    overrides work
    
    It expands upon that work by introducing many of the same concepts
    into nova.  It differs in that a few concepts were changed.  Namely
    the helm-toolkit/_oslo.tpl concept was abandoned, as there are too
    many unique oslo paths for configuration elements, so dynamic oslo
    setting mechanism was moved locally into configmap-etc.yaml, where
    it is generally used.  The references in keystone were updated and the 
    helm-toolkit function removed.
    
    Secondly, in nova the mount override effort was altered to instead
    provide mount "append" capability.  Keeping the manifest mounts
    closer to the actual pod, and keeping values.yaml a tad smaller.
    This still allows the original flexibility of operators adding
    mounts, but they cannot "undo" any of the charts built in mounts.
    
    There were several manual tweaks that needed to be done to _nova.conf.tpl
    after the oslo-config-gen process this time.  This is because there is a
    bug in the newton version of generating this file. It does not include
    the proper service authentication credentials in the [keystone_authtoken]
    section.  Future releases have a separate service credentials section
    so future config generations will likely not need these added lines.

This differs, somewhat from the keystone override pull request.  A minor subsequent PR to keystone would follow to bring that in line.

@v1k0d3n @wilreichert @DTadrzak @intlabs @wilkers-steve @larryrensing


v1k0d3n:
relates back to issue #200 (for tracking). also, @alanmeadows sorry about the conflicts.

nova/templates/daemonset-libvirt.yaml
```diff
@@ -54,54 +56,35 @@ spec:
           command:
             - bash
             - /tmp/libvirt.sh
-          volumeMounts:
-            - name: libvirtdconf
-              mountPath: /etc/libvirt/libvirtd.conf
-              subPath: libvirtd.conf
-            - name: libvirtsh
```
larryrensing: It looks like this startup script was omitted from the deployment, will the pod start?

nova/templates/daemonset-libvirt.yaml
```diff
@@ -54,54 +56,35 @@ spec:
           command:
             - bash
             - /tmp/libvirt.sh
-          volumeMounts:
-            - name: libvirtdconf
```
larryrensing: I think this should here -- was it intentionally removed?

intlabs:
@alanmeadows I've rebased this PR against master and made the changes suggested in this review as well as explicitly setting the read-only flag for mounts where appropriate in this branch: https://github.com/portdirect/openstack-helm/tree/alanmeadows-nova_config, if possible could you either apply these to this PR, or can we merge this instead?

alanmeadows:
@intlabs @larryrensing resolved, and feedback incorporated

nova/templates/daemonset-libvirt.yaml
```diff
@@ -55,53 +57,42 @@ spec:
             - bash
             - /tmp/libvirt.sh
           volumeMounts:
-            - name: libvirtdconf
+            - name: nova-etc
               mountPath: /etc/libvirt/libvirtd.conf
               subPath: libvirtd.conf
-            - name: libvirtsh
+            - name: nova-etc
```
larryrensing: this should be nova-bin

wilkers-steve:
Once the volume mount name @larryrensing pointed out gets resolved, this should be good to go

Issue #321, opened 2017-04-03
feat: Support Cinder multibackend configs.


dulek:
**Is this a bug report or feature request?** (choose one): feature

A single cinder-volume service is able to handle multiple volume backends. To make sure openstack-helm can be used with any Cinder-supported storage hardware/driver we should extend Cinder's charm to handle configuring any backend.

v1k0d3n:
this the same as #321? @wilkers-steve has a PR in currently that will bring LVM (#300), and this will ultimately take advantage of the work that @alanmeadows has done in #262, which I mentioned in your recent Cinder PR.

dulek:
Thanks, I'll make sure this have the shape to fit @wilkers-steve and @alanmeadows efforts.

PR #322, opened 2017-04-03
feat: WiP: Add support for Cinder multibackend (#321)

dulek:
A single cinder-volume service is able to handle more than one volume
backend. This commit extends _cinder.conf.tpl to be able to handle
setting multiple backends. As each of Cinder volume drivers can define
its own configuration options, the cinder/values.yaml is made more
generic and will write any option set for a certain backend directly
into cinder.conf.

### DIFF
--- a/cinder/templates/etc/_cinder.conf.tpl
+++ b/cinder/templates/etc/_cinder.conf.tpl
@@ -29,7 +29,7 @@ api_paste_config = /etc/cinder/api-paste.ini
 glance_api_servers = {{ tuple "image" "internal" "api" . | include "helm-toolkit.keystone_endpoint_uri_lookup" }}
 glance_api_version = {{ .Values.glance.version }}
 
-enabled_backends = {{  include "helm-toolkit.joinListWithComma" .Values.backends.enabled }}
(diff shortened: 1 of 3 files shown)


v1k0d3n:
this will ultimately clash with #262, which is to enable ALL of the Openstack service options. #262 is going to be a large fundamental change in the way that we deliver Value overrides.

dulek:
Thanks, I'll take a look on those and will update this PR if needed. For now I'll mark it WiP.

v1k0d3n:
sounds good @dulek. look for a PR soon coming from @alanmeadows, which will take advantage of his work here: https://github.com/alanmeadows/gen-oslo-openstack-helm.

The `gen-oslo-openstack-helm` tool can be used to generated an *entire* "out of the box" `keystone.conf`, `policy.json`, etc that can then be overridden via Values. It's really good work, but I know @alanmeadows  been tweaking it to work on things like Nova/Cinder which have backends (these complicate the use of the script). He's recently been talking about these changes/issues in the IRC channel, if you hang out there at all.

v1k0d3n:

PR #337, opened 2017-04-10
Add default dns server in dhcp agent conf.

akorzeni:
When running simple VXLAN network, VMs does not have access to DNS.
The dhcp agent pod configuration is pointing to k8s dns, but dnsmasq
agent does not have access to it within its namespace. Due to security
reasons, it should not have access to k8s underlay infra, so adding
generic 8.8.8.8 DNS server address.

Add default DNS server for simple DHCP agent configuration. 

When installing OpenStack-Helm out of the box, creating the simple VXLAN network without specifying DNS server while creating subnet, VMs launched would not have access to the DNS server.

Adding hardcoded 8.8.8.8 to the DHCP agent config. It maybe good to add it as config option.


v1k0d3n:

